PR #1, opened 2016-01-19
Few initial adjustements

ldoktor:
Let's sync the "basic" files to other avocado projects and add link to main avocado project.


apahim:
Ack. Thank you.

PR #5, opened 2016-04-05
Fix stress test after get_data_path() removal

apahim:
get_data_path() was removed from Avocado API. Also,
a new asset fetcher API was introduced in Avocado.

Given those changes, this patch fixes #4 removing
the need of get_data_path() in stress test and also
takes advantage of the asset fetcher to drop the
tarball from our repository.

Signed-off-by: Amador Pahim apahim@redhat.com

PR #6, opened 2016-04-05
Fix tests after get_data_path() removal [v2]

apahim:
v1: #5 
- Fix `stress` test, dropping `get_data_path()`.

v2:
- Fix all all occurrences of `get_data_path()` in the miscs tests repository.


clebergnu:
Looks good! Let me run them and see how they perform.

perf/aiostress.py
```diff
@@ -20,12 +20,12 @@ def setUp(self):
...
-        aiostress_c = self.params.get('aiostress_c', default='aio-stress.c')
-        c_path = self.get_data_path(aiostress_c)
-        shutil.copy(c_path, self.srcdir)
+        aiostress = self.fetch_asset('https://oss.oracle.com/~mason/aio-stress/aio-stress.c')
+        shutil.copy(aiostress, self.srcdir)
         os.chdir(self.srcdir)
         # This requires libaio.h in order to build
-        process.run('gcc -Wall -laio -lpthread -o aio-stress %s' % aiostress_c)
+        process.run('gcc -Wall -laio -lpthread -o aio-stress %s' %
+                    os.path.basename(aiostress))
```
clebergnu, line 28: I was suspicious of how/why `os.path.basename` was being used here. Turns out it made the test fail. If you think about it, the cached asset will not be in `self.srcdir`. The fix:

``` diff
diff --git a/perf/aiostress.py b/perf/aiostress.py
index 9229171..6217200 100755
--- a/perf/aiostress.py
+++ b/perf/aiostress.py
@@ -24,8 +24,7 @@ class Aiostress(Test):
         shutil.copy(aiostress, self.srcdir)
         os.chdir(self.srcdir)
         # This requires libaio.h in order to build
-        process.run('gcc -Wall -laio -lpthread -o aio-stress %s' %
-                    os.path.basename(aiostress))
+        process.run('gcc -Wall -laio -lpthread -o aio-stress %s' % aiostress)

     def test(self):
         """
```

apahim, line 28: It will be there due to the `shutil.copy(aiostress, self.srcdir)` in line 24. Anyway, I agree with your suggestion, but I'm intrigued with your failure.

clebergnu:
OK, only found that one issue.

perf/aiostress.py
clebergnu, line 28: Right, I missed that line. Anyway, no need to copy it, and yes, it failed on me.

apahim:
v3 on the way.

Issue #8, opened 2016-04-11
client tests


kalshett:
The readme says  tests ported from autotest client tests repository, so if that is the case can we name as client tests instead of Misc Tests?   

All that I'm looking for exact naming convention maintained in the autotest client tests to be available in Avocado framework. This would help us in moving from autotest -> avocado framework.

clebergnu:
@kalshett Avocado doesn't make a distinction between client and server tests, so that's why we have named this repository simply "misc" tests. You could simply drop that identification.

ldoktor:
Hello @kalshett , it's worth mention, that in the readme we say:

_This repository is dedicated to host any tests written using the avocado [1] API. It is being initially populated with tests ported from autotest client tests repository, but it's not limited by that._

so the purpose of this repository is not to port autotest tests, but to provide a share repo for various useful tests.

kalshett:
Somehow "misc" naming seems to me as uncategorised. 

I'm primarily looking at linux operating system(OS) tests to be added as part of avocado framework, if that is the case then we can categorize as OS sub-components tests?(ex: mem, filesystem(fs), IO, scheduler etc)  under this "misc" ? or you suggest different git repo under avocado?

sathnaga:
Sample directory structure can be like below?
.
├── cpu
│       ├── cpu_hotplug.py
│       └── cpu_hotplug.py.data
├── generic
├── io
│      ├── disk
│      │   ├── parallel_dd.py
│      │   └── parallel_dd.py.data
│      └── net
│             ├── netperf.py
│             └── netperf.py.data
├── mem
│       ├── memory_api.py
│       └── memory_api.py.data
└── perf

kalshett:
+1;   I'm assuming these would be under "avocado-misc-tests" repo.

apahim:
Yep, agreed.

apahim:
Feel free to reclassify tests under `perf`. Also, notice we have now an asset fetcher, so no need to keep in the repository the downloadable files, like tarballs and so on:
http://avocado-framework.readthedocs.org/en/latest/WritingTests.html#fetching-asset-files

kalshett:
yes, little bit of re-clarification is required for "perf". As not sure we really want to include
stress tests under "perf" bucket.

ldoktor:
This repository is frozen, please re-raise this issue in https://github.com/avocado-framework-tests/avocado-misc-tests if still valid.

PR #10, opened 2016-04-24
Ported error_cleanup from autotest-client-tests to avocado

balamuruhans:
Have ported error_cleanup from autotest-client-tests to avocado

### DIFF
--- a/perf/error_cleanup.py
+++ b/perf/error_cleanup.py
@@ -1,0 +1,20 @@
+#! /usr/bin/env python
+
+from avocado import Test
+from avocado import main
+
+
+class error_cleanup(Test):
+    """
+    Raise an exception during cleanup()
+    """
+
+    def test(self):
+        pass
+
+    def cleanup(self):
+        self.NameError("test a bug in cleanup()")
+
+
+if __name__ == "__main__":
+    main()


perf/error_cleanup.py
```diff
@@ -0,0 +1,20 @@
...
+
+class error_cleanup(Test):
+    """
+    Raise an exception during cleanup()
+    """
+
+    def test(self):
+        pass
+
+    def cleanup(self):
```
apahim, line 15: In avocado we don't have `cleanup()`, but `tearDown()`.

apahim:
Wondering what's the purpose of this test? Seems like a selftest for Avocado and if so, it should sit on avocado main repo, under `selftest/functional` directory.

balamuruhans:
@apahim I tried to port simple test available from autotest to learn how avocado works. So, can change it with tearDown() and put it in selftest/functional or it is not required to be added?

apahim:
@balamuruhans if you just change from `cleanup()` to `tearDown()` then you will have successfully ported the test. The `selftest/functional` part I mentioned is a different discussion, intended to understand the use and the relevance of the test. Also, since this is not a `perf` test for sure, you can create a new directory for it, like `generic`, as discussed here: https://github.com/avocado-framework/avocado-misc-tests/issues/8.

balamuruhans:
@apahim thank you, I will add new directory as generic and add it.

PR #15, opened 2016-04-26
Dbench Test for Avocado

harish-24:
Ported Dbench Test from autotest-client-test to avocado-misc-tests

### DIFF
--- a/io/disk/dbench.py
+++ b/io/disk/dbench.py
@@ -1,0 +1,71 @@
+import os
+import re
+import multiprocessing
+import json
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import process
+from avocado.utils import build
+
+
+class Dbench(Test):
+
+    """
+    Dbench is a tool to generate I/O workloads to either a filesystem or to a
+    networked CIFS or NFS server.
+    Dbench is a utility to benchmark a system based on client workload
+    profiles.
+    """
+
+    def setUp(self):
+        '''
+        Build Dbench
+        Source:
+        http://samba.org/ftp/tridge/dbench/dbench-3.04.tar.gz
+        '''
+
+        try:
+            process.run('which gcc')
+        except:
+            self.error('gcc is required by this job and is not available on'
+                       'the system')
+
+        self.results = []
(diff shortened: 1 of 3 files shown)


io/disk/dbench.py
```diff
@@ -0,0 +1,71 @@
...
+        self.results = []
+        data_dir = os.path.abspath(self.datadir)
+        tarball = self.fetch_asset(
+            'http://samba.org/ftp/tridge/dbench/dbench-3.04.tar.gz')
+        archive.extract(tarball, self.srcdir)
+        cb_version = os.path.basename(tarball.split('.tar.')[0])
+        self.srcdir = os.path.join(self.srcdir, cb_version)
+        os.chdir(self.srcdir)
+        patch = self.params.get('patch', default='dbench_startup.patch')
+        process.system('patch -p1 < %s' % data_dir + '/' + patch, shell=True)
```
apahim: The practical difference between `process.system()` and `process.run()` is the first returns the command exit code and the last returns the `CmdResult` class object.
Since you're not using the functions returns, you could use the same function for both cases so you don't confuse code readers.

apahim:
@harish-24 thank you for this port. I have a very minor comment. Please tell me if you agree in adjust or if you prefer to keep this version and then we are good to go.

harish-24:
@apahim I agree. I have pushed a new commit with the change mentioned.

io/disk/dbench.py
```diff
@@ -0,0 +1,71 @@
...
+        cmd = '%s/dbench %s %s -D %s -c %s -t %d' % (self.srcdir, nprocs, args,
+                                                     dir, loadfile, seconds)
+        process.run(cmd)
+
+        self.results = process.system_output(cmd)
+        pattern = re.compile(r"Throughput (.*?) MB/sec (.*?) procs")
+        (throughput, procs) = pattern.findall(self.results)[0]
+        perf_json = {'throughput': throughput, 'procs': procs}
+        output_path = os.path.join(self.outputdir, "perf.json")
+        json.dump(perf_json, open(output_path, "w"))
```
apahim, line 68: Nice!

apahim:
Looks good. Thank you.

PR #16, opened 2016-04-28
Installing dependencies for Dbench Test

harish-24:
As suggested, SoftwareManager is used for installing dependency

### DIFF
--- a/io/disk/dbench.py
+++ b/io/disk/dbench.py
@@ -8,6 +8,7 @@ from avocado import main
 from avocado.utils import archive
 from avocado.utils import process
 from avocado.utils import build
+from avocado.utils.software_manager import SoftwareManager
 
 
 class Dbench(Test):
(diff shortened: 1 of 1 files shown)


apahim:
There are 2 commits in this pull request. You should probably smash them or remove the first.

harish-24:
I have pushed as a single patch. Thanks

io/disk/dbench.py
```diff
@@ -25,12 +26,9 @@ def setUp(self):
...
         '''
-
-        try:
-            process.run('which gcc')
-        except:
-            self.error('gcc is required by this job and is not available on'
-                       'the system')
+        sm = SoftwareManager()
+        if not sm.check_installed("gcc"):
+            sm.install("gcc")
```
apahim: Will you continue the test on install failure?

apahim: `install()` returns `False` on install failure. You can just:

```
...
        if not sm.check_installed("gcc") and not sm.install("gcc"):
            self.error('need gcc')
...
```

harish-24:
@apahim  Agreed . Have sent as a new commit.

apahim:
Looks good to me. Thank you.

PR #24, opened 2016-05-16
Porting ltrace test from autotest-client-tests to avocado-misc-tests

jayarama-ibm:


toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
```
ldoktor, line 1: Hello @jayarama-ibm would you please add the license heading? You can inspire here: https://github.com/avocado-framework/avocado-misc-tests/pull/26

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import process
+from avocado.utils import git
+from avocado.utils import distro
+from avocado.utils.software_manager import SoftwareManager
+
+backend = SoftwareManager()
+dist = distro.detect()
```
ldoktor: Top level variables should use UPPER_CASE only. But I see you use it only inside the `setUp` method, would you please move it there, instead to avoid import exceptions and unnecessary memory overhead?

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+
+        os_deps = ['git', 'wget', 'rpm', 'autoconf',
+                   'automake', 'dejagnu', 'binutils']
+
+        if dist.name == 'sles':
+            sles_deps = ['build', 'libdw-devel', 'libelf-devel',
+                         'elfutils', 'binutils-devel', 'libtool', 'gcc-c++']
+            pkgs = os_deps + sles_deps
+
+        elif dist.name == 'redhat':
```
ldoktor: Would you please use `dist.name in ("redhat", "fedora")`? They should be compatible.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+            rhel_deps = ['elfutils-devel', 'elfutils-libelf-devel',
+                         'elfutils-libelf', 'elfutils-libs', 'libtool-ltdl']
+            pkgs = os_deps + rhel_deps
+
+        elif dist.name == 'ubuntu':
+            ubuntu_deps = ['elfutils', 'libelf-dev', 'libtool',
+                           'libelf1', 'librpmbuild3', 'binutils-dev']
+            pkgs = os_deps + ubuntu_deps
+        else:
+            print "Unsupported OS!"
```
ldoktor: Please don't use prints, instead use `self.log`. And for this kind of message `self.log.warn` would be IMO the most appropriate.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+
+        i = 0
+        while i < len(pkgs):
```
ldoktor: Well how about using `for pkg in pkgs:`

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+            if backend.install(pkgs[i]):
+                print "%s installed successfull" % pkgs[i]
```
ldoktor: this is quite verbose, I'd recommend to log only errors (well actually fail the test in that case). It's not strong opinion, thought so feel free to either remove it, or use `self.log.debug` instead.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+            else:
+                print "Error: Check repository !"
+                exit(1)
```
ldoktor: Please never use `exit` directly. It works, but is not really nice. The most suitable would be `self.skip`, which says the test is not available on your machine. I'd also agree with `self.error` which says unrelated error occurred.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+            i = i + 1
+
+        """
+        clone the  ltrace tests from the GIT repo
+        Source: git clone git://git.debian.org/git/collab-maint/ltrace.git
+        """
```
ldoktor: I don't think this comment is necessary. If you think it is, please use `#` instead of `"""`. The `"""` is reserved for variables (and docstrings).

toolchain/ltrace.py
ldoktor, line 1: Btw @jayarama-ibm I'm searching the test in `autotest-client-tests` and I can't find it there. So I guess it was internal one, right?

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+
+        git.get_repo('git://git.debian.org/git/collab-maint/ltrace.git',
+                     destination_dir=self.srcdir + '/ltrace')
```
ldoktor: please use `os.path.join`

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+        ltrace_version = os.path.basename("ltrace")
+        os.chdir(self.srcdir)
+
+        self.srcdir = os.path.join(self.srcdir, ltrace_version)
+        os.chdir(self.srcdir)
+        process.run('./autogen.sh')
+        process.run('./configure')
+        build.make(self.srcdir)
+        try:
+            process.run('which gcc')
```
ldoktor: This is not necessary as `./configure` fails when there is no `gcc`.

toolchain/ltrace.py
ldoktor: Also in case this branch is used, the pkgs is never set. I'd suggest using `pkgs` from the beginning and extend it of the dist-dependant versions by `pkgs += rhel_deps` and similarly for others.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+            i = i + 1
+
+        """
+        clone the  ltrace tests from the GIT repo
+        Source: git clone git://git.debian.org/git/collab-maint/ltrace.git
+        """
+
+        git.get_repo('git://git.debian.org/git/collab-maint/ltrace.git',
+                     destination_dir=self.srcdir + '/ltrace')
+        ltrace_version = os.path.basename("ltrace")
```
ldoktor: I don't see a reason for calling `os.path.basename`. You chosen the destination dir, it's always `ltrace`. It has nothing to do with ltrace_version.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+        os.chdir(self.srcdir)
+
+        self.srcdir = os.path.join(self.srcdir, ltrace_version)
```
ldoktor: It's not really nice to change the `self.srcdir`. It's assigned by the runner. Would you mind renaming it to eg. `self.ltrace_git` or `self.ltrace_src`?

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+
+        """
+        clone the  ltrace tests from the GIT repo
+        Source: git clone git://git.debian.org/git/collab-maint/ltrace.git
+        """
+
+        git.get_repo('git://git.debian.org/git/collab-maint/ltrace.git',
+                     destination_dir=self.srcdir + '/ltrace')
+        ltrace_version = os.path.basename("ltrace")
+        os.chdir(self.srcdir)
```
ldoktor: There is no need to change dir here as you're changing it once again in +3 lines.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+        process.run('./configure')
+        build.make(self.srcdir)
+        try:
+            process.run('which gcc')
+        except error.CmdError:
+            raise NotAvailableError('gcc is required by this job and is '
+                                    'not available on the system')
+
+    def test(self):
+        os.chdir(self.srcdir)
```
ldoktor: This is not necessary. First you already do that in setUp and second the `build.make` also does that.

toolchain/ltrace.py
```diff
@@ -0,0 +1,82 @@
...
+        build.make(self.srcdir, extra_args='check')
```
ldoktor: Basically the output of this test is similar to binutils, so let me paste the suggestion here:

This runs the checks and produces huge output, but one does not see the details. So basically this test gives you the PASS/FAIL, but later on you can't tell why it failed. Would you please copy the `*.sum` files from the `binutils` build directory to the test results to allow postprocess?

Additionally it would be convenient to list the failed cases if it fails. Manually I do `for summary in $(find . -name '*.sum'); do echo $summary; grep ^FAIL $summary; done` in the build directory.

So I'd suggest:

``` python
for root, dirnames, filenames in os.walk('.'):
    for filename in fnmatch.filter(filenames, '*.sum'):
        filename = os.path.join(root, filename)
        logfile = filename[:-4] + ".log"
        # copy logfile to self.logdir
        # open(filename); read line-by-line and self.log.error lines starting with `FAIL`
        # (alternatively attach the full filename as it might be useful to see what tests were skipped)
```

What do you think? (I'm not enforcing you to do that, I just think it'd be useful to attach those files).

toolchain/ltrace.py
jayarama-ibm, line 1: ltrace is is the internal test case. I am working on the review comments.

toolchain/ltrace.py
jayarama-ibm: Hi ldoktor

I have incorporated all the review comments which are  mentioned.
Ran the ltrace.py through avocado.
Observations.
While  performing  the "make check" - it stays there in make check for longer duration - It is not coming out from this loop.
After that added the code to check the *.log files.
It is not entering the next lines of code to check .
How to go about this statements to execute and make the test is pass/fail . Please suggest.

ldoktor: Hello @jayarama-ibm, I'm not sure I follow, would you please push the new version so I can see the problem?

jayarama-ibm:
Attached the new  "ltrace.py" where all the review comments are incorporated.

toolchain/ltrace.py
```diff
@@ -0,0 +1,104 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM.
+# Author: Jayarama<jsomalra@linux.vnet.ibm.com>
+
+# Based on code by
+#   Author: Abdul Haleem <abdhalee@linux.vnet.ibm.com>
```
ldoktor, line 17: What was the license and copyright of the original code? Where can it be found? Do you have permission from Abdul to re-use and modify it?

toolchain/ltrace.py
```diff
@@ -0,0 +1,104 @@
...
+    def setUp(self):
+
+        backend = SoftwareManager()
+        dist = distro.detect()
+
+        """
+        Install  all the required dependencies
+        Building source tarball requires packages specific to os
+        that needs to be installed, if not installed test will stop.
+        """
```
ldoktor, line 42: Docstring have to be first in order to work as the function docstring.

toolchain/ltrace.py
```diff
@@ -0,0 +1,104 @@
...
+
+        pkgs = ['git', 'wget', 'rpm', 'autoconf', 'automake',
+                'dejagnu', 'binutils']
+
+        if dist.name == 'sles':
+            sles_deps = ['build', 'libdw-devel', 'libelf-devel',
+                         'elfutils', 'binutils-devel', 'libtool', 'gcc-c++']
+            pkgs += sles_deps
+
+        elif dist.name == ('redhat'):
```
ldoktor: The parenthesis are not necessary. My recommendation was (and still is) to use:

```
    ...
elif dist.name in ("redhat", "fedora"):
    ...
elif dist.name == "ubuntu":
    ...
```

Parenthesis there would be necessary as you'd be looking for the `dist.name` in tuple.

toolchain/ltrace.py
```diff
@@ -0,0 +1,104 @@
...
+                           'libelf1', 'librpmbuild3', 'binutils-dev']
+            pkgs += ubuntu_deps
+        else:
+            self.log.warn("Unsupported OS!")
+
+        for pkg in pkgs:
+            if backend.install(pkg):
+                self.log.debug("%s installed successfull" % pkg)
+            else:
+                self.error("Error: Check repository !")
```
ldoktor: I'd not like to debug such failure. How about:

``` python
for pkg in pkgs:
    if backend.check_installed(pkg) and not backend.install(pkg):
        self.error("Fail to install package %s required for this test.")
```

ldoktor: If you insist on logging the successfully installed packages, then:

``` python
for pkg in pkgs:
    if backend.check_installed(pkg):
        if backend.install(pkg):
            self.log("%s installed successfully", pkg)
        else:
            self.error("Fail to install package %s required for this test." % pkg)
```

toolchain/ltrace.py
```diff
@@ -0,0 +1,104 @@
...
+        process.run('./autogen.sh')
+        process.run('./configure')
+        build.make(self.src_lt)
+
+    def test(self):
+        build.make(self.src_lt, extra_args='check')
+        for root, dirnames, filenames in os.walk('.'):
+            for filename in fnmatch.filter(filenames, '*.log'):
+                filename = os.path.join(root, filename)
+                logfile = filename[:-4] + ".log"
```
ldoktor: OK I don't understand the need to remove `.log` and then add it back... Without this line it should work the same, shouldn't it?

toolchain/ltrace.py
```diff
@@ -0,0 +1,104 @@
...
+                os.system('cp ' + logfile + ' ' + self.logdir)
```
ldoktor: Please don't use `cp` to copy files, use `shutil.copy` instead.

jayarama-ibm:
ldoktor,
I am working to fix the  review comments and send the updated version.

ldoktor:
OK, please send a v2 as new PR along with changelog. Also does this test requires the latest version? It'd be way faster to use `self.fetch_asset("http://www.ltrace.org/ltrace_0.7.3.orig.tar.bz2")` instead and we could update the link when new version occurs...

toolchain/ltrace.py
jayarama-ibm, line 17: yes

jayarama-ibm:
Hi Idoktor,
I downloaded "http://www.ltrace.org/ltrace_0.7.3.orig.tar.bz2" to check this ltrace.
There are some while running ./configure --> for this updated the config.guess able to run "./configure"

The following issues are faced while running the make  ==>
make[2]: **\* [ltrace-elf.lo] Error 1
make[2]: Leaving directory '/root/sj1/avocado1/avocado-fvt-wrapper/avocado-misc-tests/toolchain/toolchain.py.data/http-ltrace/ltrace-0.7.3'
Makefile:741: recipe for target 'all-recursive' failed
make[1]: **\* [all-recursive] Error 1
make[1]: Leaving directory '/root/sj1/avocado1/avocado-fvt-wrapper/avocado-misc-tests/toolchain/toolchain.py.data/http-ltrace/ltrace-0.7.3'
Makefile:447: recipe for target 'all' failed
make: **\* [all] Error 2
For this 3 files are modified ==> Able to do make

> > Then ran the "make check" to run the ltrace tests ==> It is failing at several places and producing the proper results.

Test output:
make[2]: **\* [ltrace-elf.lo] Error 1
make[2]: Leaving directory '/root/sj1/avocado1/avocado-fvt-wrapper/avocado-misc-tests/toolchain/toolchain.py.data/http-ltrace/ltrace-0.7.3'
Makefile:741: recipe for target 'all-recursive' failed
make[1]: **\* [all-recursive] Error 1
make[1]: Leaving directory '/root/sj1/avocado1/avocado-fvt-wrapper/avocado-misc-tests/toolchain/toolchain.py.data/http-ltrace/ltrace-0.7.3'
Makefile:447: recipe for target 'all' failed
make: **\* [all] Error 2

Note:
The "git clone" of ltrace able to run
./autogen.sh
configure, 
make -- running properly

make check ==> Giving some failures.

Suggest how to go about this.

ldoktor:
I see, then please use the git for now and we can change it once we have compilable version...

jayarama-ibm:
Hi Idoktor,
The review comments are added for ltrace test. Please check and push the patch.

ldoktor:
Hello @jayarama-ibm, it's almost there, I spotted few little issues, please take a look at #121.

PR #25, opened 2016-05-16
Binutils test for avocado

jayarama-ibm:


toolchain/binutils.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
```
ldoktor, line 1: Hello @jayarama-ibm, I haven't found this in Autotest, so did you implement this test, or is it port from other test-framework? I'm asking because there needs to be the license header with author and/or copyright. So please add GPL compatible header and the author (either you or link to the original source).

Please see https://github.com/avocado-framework/avocado/blob/master/avocado/utils/software_manager.py for details.

toolchain/binutils.py
```diff
@@ -0,0 +1,53 @@
...
+        backend.install('binutils-devel')
+        backend.install('dejagnu')
+        backend.install('libtool')
+
+        """
+        Extract - binutils
+        Source: http://ftp.gnu.org/gnu/binutils/binutils-2.25.1.tar.bz2
+        """
+        tarball = self.fetch_asset(
+            'http://ftp.gnu.org/gnu/binutils/binutils-2.25.1.tar.bz2')
```
ldoktor: They do support https, would you mind using it instead?

toolchain/binutils.py
```diff
@@ -0,0 +1,53 @@
...
+        backend.install('autoconf')
+        backend.install('automake')
+        backend.install('binutils-devel')
+        backend.install('dejagnu')
+        backend.install('libtool')
+
+        """
+        Extract - binutils
+        Source: http://ftp.gnu.org/gnu/binutils/binutils-2.25.1.tar.bz2
+        """
```
ldoktor: Do you think this comment is necessary? I'd either skip it or use `#` notation as `"""` are reserved for values or docstrings.

toolchain/binutils.py
```diff
@@ -0,0 +1,53 @@
...
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import build
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+backend = SoftwareManager()
```
ldoktor: Global variables should use UPPER_CASE notation. Still I think it'd be better to move this inside the `setUp` as it's only used there and having it in the main level makes it prone to import errors.

toolchain/binutils.py
```diff
@@ -0,0 +1,53 @@
...
+        Install required tools and resolve dependencies
+        '''
+        backend.install('rpmbuild')
+        backend.install('elfutils')
+        backend.install('build')
+        backend.install('autoconf')
+        backend.install('automake')
+        backend.install('binutils-devel')
+        backend.install('dejagnu')
+        backend.install('libtool')
```
ldoktor: There are couple of dependencies required on Fedora.23. Could you please add: `glibc-static` and `zlib-static`?

ldoktor: Also the `backend.install` does not raise exception when it fails to install the package. Would you please add `self.skip` with the list of packages which failed to install? If you expect some install failures (as package names varies over distros) you might either add `res = backend.install("glibc-static") or backend.install("glibc-devel-static")` and then check for res.

Alternatively if you don't see those as a big problem, please add at least `self.warn` listing the packages which failed to install. Usually the binutils tests fail due to incorrect environment.

toolchain/binutils.py
ldoktor: Also the 2.26 version is available and seems to work fine.

toolchain/binutils.py
```diff
@@ -0,0 +1,53 @@
...
+
+        bintools_version = os.path.basename(tarball.split('.tar.')[0])
+        self.srcdir = os.path.join(self.srcdir, bintools_version)
+
+    def test(self):
+        os.chdir(self.srcdir)
+
+        process.run('./configure')
+        build.make(self.srcdir)
+        build.make(self.srcdir, extra_args='check')
```
ldoktor: This runs the checks and produces huge output, but one does not see the details. So basically this test gives you the PASS/FAIL, but later on you can't tell why it failed. Would you please copy the `*.sum` files from the `binutils` build directory to the test results to allow postprocess?

Additionally it would be convenient to list the failed cases if it fails. Manually I do `for summary in $(find . -name '*.sum'); do echo $summary; grep ^FAIL $summary; done` in the build directory.

So I'd suggest:

``` python
for root, dirnames, filenames in os.walk('.'):
    for filename in fnmatch.filter(filenames, '*.sum'):
        filename = os.path.join(root, filename)
        logfile = filename[:-4] + ".log"
        # copy logfile to self.logdir
        # open(filename); read line-by-line and self.log.error lines starting with `FAIL`
        # (alternatively attach the full filename as it might be useful to see what tests were skipped)
```

What do you think?

toolchain/binutils.py
ldoktor, line 1: Actually you can inspire here: https://github.com/avocado-framework/avocado-misc-tests/pull/26 for heading (both versions are acceptable)

jayarama-ibm:
binutils is the new test developed.
I will work on these comments and update  the status.

ldoktor:
Cool, then please add GPL compatible header (ideally GPLv2+) including the copyright/author and please take care (or feel free to disagree with some) of the issues.

toolchain/binutils.py
jayarama-ibm: I am not seeing the multiple *.sum files. It  is generating one .sum file with name "md5.sum" in binutils-2.26 directory. Is that fine to handle this file ?

ldoktor: Oh my fault, it's not `*.sum`, but `*.log`.

jayarama-ibm:
Hi Idoktor,
The review comments are added for binutils test. Please check and push the patch.

ldoktor:
Hello @jayarama-ibm, thank you for the updates, it looks almost good for inclusion, there are few issues, please take a look at #120. I'll merge it if you agree with the changes...

jayarama-ibm:
Hello Idoktor
I checked the code. It looks fine.Please merge with the changes.

PR #27, opened 2016-05-19
Adding test for Module Unload/Load

hathyaga:
This Program loads and unloads the kernel driver modules along with the dependant modules for a total of 100 counts.

### DIFF
--- a/io/driver/module_unload_load.sh
+++ b/io/driver/module_unload_load.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+CONFIG_FILE="$AVOCADO_TEST_DATADIR"/config
+DRIVER=`lspci -k | grep -iw "Kernel driver in use" | cut -d ':' -f2 | sort | uniq`
+
+#Loads the kernel driver modules
+module_load() {
+    echo "Reloading driver $1"
+    modprobe $1
+    if [ $? != 0 ]; then
+        echo "Failed to load driver module $1"
+        break;
+    fi
+    echo
+}
+
+#Unloads the kernel driver modules
+module_unload() {
+    for i in $(cat $CONFIG_FILE | grep "$1=" | awk -F'=' '{print $2}'); do
+        module_unload $i
+        if [ $? != 0 ]; then
+            echo "Failed to unload driver module $i"
+            return
+        fi
+    done
+    echo "Unloaded driver $1"
+    rmmod $1
+    return
+}
+
+#Parses the input and then performs module load/unload 
+for driver in $DRIVER; do
+    echo "Starting driver module load/unload test for $driver"
+    echo
+    for j in $(seq 1 100) ; do
+            module_unload $driver
+            # Sleep for 5s to allow the module load to complete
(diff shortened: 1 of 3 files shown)


io/driver/module_unload_load.sh
```diff
@@ -0,0 +1,42 @@
...
+module_unload() {
+    for i in $(cat $CONFIG_FILE | grep "$1=" | awk -F'=' '{print $2}'); do
+        module_unload $i
+        if [ $? != 0 ]; then
+            echo "Failed to unload driver module $i"
+            return
+        fi
+    done
+    echo "Unloaded driver $1"
+    rmmod $1
```
ldoktor, line 26: There is no check for return number so I'm wondering whether this should be called with `-e`, or if the purpose of this test is not to remove the modules, but only to try to do it and see if the system crashes or not, right? If the later, then avocado has no support for reboot (kernel panics) during execution so the user would only get the partial test results stored in the results dir, without the `json`, `xunit`/... results (basically no results for non-streamlined output plugins).

Would you please clarify what is the purpose of this test?

io/driver/module_unload_load.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
ldoktor, line 1: Would you please start with the license heading? You can inspire from other tests ins `avocado-misc-test` repo.

io/driver/module_unload_load.sh
ldoktor, line 26: Btw on F23 running inside VM this test effectively hangs the VM by removing the virtio-pci, which means there is no disk attached. I'm really wondering what the purpose of this test is and I think it should require a parameter what drivers it should try adding/removing using some sane default like `scsi_debug`.

hathyaga, line 26: The purpose is to unload and load all the kernel loaded modules for 100 times. 
Earlier, the code did not have the functionality of what to do when the module is not found. It now has been changed to handle this

Using "-e " is not feasible here. When used with -e, the code exits if a module is not found and this is not the expected behaviour. The code needs to skip the execution when the module is not found and continue with the rest of them.

hathyaga, line 26: As for virtio-pci, I ran the code on a PowerKVM guest and it worked fine. Please can you let me know on which system and where did it hang.

ldoktor, line 26: Yep, I understand the point, but in reality the modules might not be unloaded as the `rmmod` could fail and the test won't complain about it so I'd prefer checking the error and failing the test in such case. But if you consider this test only as stress test, then fine, but it should be mentioned in the heading.

About the system, I simply run Fedora 23 on Fedora 23, x86_64, no special handling, but the failure is not important, the test behaved correctly. We might later expand some blacklist to avoid such problems, but it should be based on real-world usage, not just me trying the test.

hathyaga:
Closing this pull request. I will be generating a new pull request containing code having taken care of all the suggestions mentioned here.

PR #26, opened 2016-05-18
Add license headings

ldoktor:
This patch adds license headings and it will be merged only when all
authors agrees with the change.

Signed-off-by: Lukáš Doktor ldoktor@redhat.com

PR #28, opened 2016-05-19
Ported disktest from Autotest

Nosheen08:
Ported disktest from autotest to avocado.

### DIFF
--- a/io/disk/disktest.py
+++ b/io/disk/disktest.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation; either version 2 of the License, or
@@ -9,10 +10,23 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 #
 # See LICENSE for more details.
+#
 # Copyright: 2016 IBM
 # Author: Nosheen Pathan <nopathan@linux.vnet.ibm.com>
 # Based on code by
-# Author: Martin Bligh (mbligh@google.com)
+
+#   Author: Lucas Meneghel Rodrigues <lmr@redhat.com>
+#   Author: Yiqiao Pu <ypu@redhat.com>
+#   Author: Jiří Župka <jzupka@redhat.com>
+#   Author: Martin J. Bligh <mbligh@google.com>
+#   Author: John Admanski <jadmanski@google.com>
+#   Author: Mihai Rusu <dizzy@google.com>
+#   Copyright: 2009-2015  Red Hat, Inc
+#   Copyright: 2013  Red Hat, Inc
+#   Copyright: 2012  Red Hat, Inc
+#   Copyright: 2010  Google, Inc
+#   Copyright: 2010  Google, Inc
+#   Copyright: 2009 Google, Inc
 
 import os
 import shutil
--- a/io/disk/disktest.py
+++ b/io/disk/disktest.py
@@ -13,20 +13,22 @@
 #
 # Copyright: 2016 IBM
 # Author: Nosheen Pathan <nopathan@linux.vnet.ibm.com>
-# Based on code by
 
+# Based on code by
 #   Author: Lucas Meneghel Rodrigues <lmr@redhat.com>
 #   Author: Yiqiao Pu <ypu@redhat.com>
 #   Author: Jiří Župka <jzupka@redhat.com>
 #   Author: Martin J. Bligh <mbligh@google.com>
 #   Author: John Admanski <jadmanski@google.com>
 #   Author: Mihai Rusu <dizzy@google.com>
+#   Author: Scott Zawalski <scottz@google.com>
 #   Copyright: 2009-2015  Red Hat, Inc
 #   Copyright: 2013  Red Hat, Inc
 #   Copyright: 2012  Red Hat, Inc
-#   Copyright: 2010  Google, Inc
+#   Copyright: 2007-2010  Google, Inc
 #   Copyright: 2010  Google, Inc
 #   Copyright: 2009 Google, Inc
+#   Copyright: 2008 Google, Inc
 
 import os
 import shutil
@@ -135,8 +137,6 @@ class Disktest(Test):
                 errors.append(retval)
         if errors:
             self.fail("Errors from children: %s" % errors)
-
-        # self.tearDown()
 
     def tearDown(self):
         """
--- a/io/disk/disktest.py
+++ b/io/disk/disktest.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+# Copyright: 2016 IBM
+# Author: Nosheen Pathan <nopathan@linux.vnet.ibm.com>
+# Based on code by
+# Author: Martin Bligh (mbligh@google.com)
+
+import os
+import shutil
+
+from avocado import Test
+from avocado import main
+from avocado.utils import memory
+from avocado.utils import disk
+from avocado.utils import process
+from avocado.utils import build
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Disktest(Test):
+
+    """
+    Avocado module for disktest.
+    Pattern test of the disk, using unique signatures for each block and each
+    iteration of the test. Designed to check for data corruption issues in the
+    disk and disk controller.
+    It writes 50MB/s of 500KB size ops.
+    """
+    version = 2
+
+    def setUp(self):
+        """
+        Verifies if we have gcc to compile disktest.
+        :param disks: Directory (usually mountpoints) to be passed to the test.
+                      Directory gets created if not existing
+        :param gigabytes: Disk space that will be used for the test to run.
+        :param chunk_mb: Size of the portion of the disk used to run the test.
+                        Cannot be larger than the total amount of free RAM.
+        :param source: name of the source file located in deps path
+        :param make: name of the makefile file located in deps path
+        """
+        softm = SoftwareManager()
+        if not softm.check_installed("gcc") and not softm.install("gcc"):
+            self.error('Gcc is needed for the test to be run')
+
+        self.disks = self.params.get('disks', default=self.workdir)
+        self.gigabytes = self.params.get('gigabytes', default=None)
+        self.chunk_mb = self.params.get('chunk_mb', default=None)
+        source = self.params.get('source', default='disktest.c')
+        makefile = self.params.get('make', default='Makefile')
+        c_file = os.path.join(self.datadir, source)
+        c_file_name = os.path.basename(c_file)
+        make_file = os.path.join(self.datadir, makefile)
+        make_file_name = os.path.basename(make_file)
+        dest_c_file = os.path.join(self.srcdir, c_file_name)
+        dest_m_file = os.path.join(self.srcdir, make_file_name)
+        shutil.copy(c_file, dest_c_file)
+        shutil.copy(make_file, dest_m_file)
+        self.memory_mb = memory.memtotal() / 1024 / 8
+        os.chdir(self.srcdir)
+        build.make(self.srcdir)
+
+    def one_disk_chunk(self, disks, chunk):
+        """
+        Tests one part of the disk by spawning a disktest instance.
+        :param disks: Directory (usually a mountpoint).
+        :param chunk: Portion of the disk used.
+        """
+        self.log.info("Testing %d MB files on %s in %d MB memory, chunk %s",
+                      self.chunk_mb, disks, self.memory_mb, chunk)
+        logfile = os.path.join(self.outputdir, "log.txt")
+        cmd = ("%s/disktest -m %d -f %s/testfile.%d -i -S >>%s 2>&1" %
+               (self.srcdir, self.chunk_mb, disks, chunk, logfile))
+        self.log.debug("Running '%s'", cmd)
+
+        proc = process.get_sub_process_klass(cmd)(cmd, shell=True)
+        proc.start()
+        proc.poll()
+        proc.wait()
+        return proc
+
+    def test(self):
+        """
+        Runs one iteration of disktest.
+
+        """
+        os.chdir(self.srcdir)
+        if self.chunk_mb is None:
+            self.chunk_mb = memory.memtotal() / 1024 / 8
+        if self.gigabytes is None:
+            free = 100  # cap it at 100GB by default
+            if not os.path.isdir(self.disks):
+                os.makedirs(self.disks)
+
+            free = min(disk.freespace(self.disks) / 1024 ** 3, free)
+            self.gigabytes = free
+            self.log.info("Resizing to %s GB", self.gigabytes)
+
+        if self.memory_mb > self.chunk_mb:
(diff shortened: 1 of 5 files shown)


io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
+import os
```
ldoktor: Hello @Nosheen08, please include the license on top of this file. You can inspire here https://github.com/avocado-framework/avocado-misc-tests/pull/26/commits/309eb378728d0fe3bfd2d0573967341a8bb07b59

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+        build.make(srcdir, extra_args='clean')
+        build.make(srcdir)
+
+    def one_disk_chunk(self, disks, chunk):
+        """
+        Tests one part of the disk by spawning a disktest instance.
+        :param disks: Directory (usually a mountpoint).
+        :param chunk: Portion of the disk used.
+        """
+        logging.info("Testing %d MB files on %s in %d MB memory, chunk %s",
```
ldoktor: Would you please use `self.log` instead of the root logger from `logging`.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+                     self.chunk_mb, disks, self.memory_mb, chunk)
+        logfile = self.outputdir + '/log.txt'
+        cmd = ("%s/disktest -m %d -f %s/testfile.%d -i -S >>%s 2>&1" %
+               (self.disk_srcdir, self.chunk_mb, disks, chunk, logfile))
+        logging.debug("Running '%s'", cmd)
+        thread = subprocess.Popen(cmd, shell=True)
```
ldoktor: avocado has a modified version for subprocesses in `avocado.utils.process`. It'd be better to use that one instead. The benefit is you can use wrappers/gdb/... dynamically from cmdline (used by developers).

``` python
proc = process.get_sub_process_klass(cmd)(cmd, shell=True)
pid = proc.start()
...
proc.poll()
result = proc.wait()
```

This way one can use features like: http://avocado-framework.readthedocs.io/en/latest/DebuggingWithGDB.html and http://avocado-framework.readthedocs.io/en/latest/WrapProcess.html

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+                       self.memory_mb)
+
+        chunks = (1024 * self.gigabytes) / self.chunk_mb
+        logging.info("Total of disk chunks that will be used: %s", chunks)
+        for i in range(chunks):
+            pids = []
+            pid = self.one_disk_chunk(self.disks, i)
+            errors = []
+            for pid in pids:
+                (junk, retval) = os.waitpid(pid, 0)
```
ldoktor: even with subprocess it's preferred to use `proc.wait()`. So the `self.one_disk_chunk` should return the `proc` instance, instead of `pid`.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+                if retval != 0:
+                    errors.append(retval)
+            if errors:
+                self.error("Errors from children: %s" % errors)
+
+        self.cleanup(self.disks)
```
ldoktor: This should be probably located in `tearDown` to make sure it's executed even on test error. (basically you could just rename the `cleanup` method and use `self.disks` directly and not as argument)

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+        Verifies if we have gcc to compile disktest.
+        """
+        softm = SoftwareManager()
+        if not softm.check_installed("gcc") and not softm.install("gcc"):
+            self.error('Gcc is needed for the test to be run')
+        self.disks = self.params.get('disks', default=None)
+        self.gigabytes = self.params.get('gigabytes', default=None)
+        self.chunk_mb = self.params.get('chunk_mb', default=None)
+        self.memory_mb = memory.memtotal() / 1024 / 8
+        self.data_dir = os.path.join(os.path.abspath(self.datadir))
```
ldoktor: I guess you forgot to put something here? Btw the `datadir` is already abspath, so you can probably directly use `self.datadir`.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+        self.disk_srcdir = os.path.join(os.path.abspath(self.datadir), 'src')
+        self.build(self.disk_srcdir)
```
ldoktor: The `datadir` might not be writable. The recommended workflow is to copy the file to `self.srcdir` and simply compile it. You don't even need the makefile as avocado has utility for that. Take a look at https://github.com/avocado-framework/avocado/blob/master/examples/tests/doublefree.py

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def build(self, srcdir):
+        """
+        Compiles disktest.
+        """
+        os.chdir(srcdir)
+        build.make(srcdir, extra_args='clean')
```
ldoktor: With the workflow mentioned above (copy the sources to `self.srcdir`) you don't need to call the `make clean` as it's a fresh tmpdir.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+                to the test. Directory gets created if not existing
+        :param gigabytes: Disk space that will be used for the test to run.
+        :param chunk_mb: Size of the portion of the disk used to run the test.
+                Cannot be larger than the total amount of free RAM.
+        """
+        os.chdir(self.disk_srcdir)
+        if self.chunk_mb is None:
+            self.chunk_mb = memory.memtotal() / 1024 / 8
+        if self.disks is None:
+            self.disks = self.data_dir
```
ldoktor: this location might not be writable. You should probably use `self.workdir` and as `srcdir` is inside of this, I'd recommend using `avocado.utils.path.init_dir(self.workdir, "testdisk")` to avoid possible problems.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+        if self.gigabytes is None:
+            free = 100  # cap it at 100GB by default
+            if not os.path.isdir(self.disks):
+                os.makedirs(self.disks)
+
+            free = min(disk.freespace(self.disks) / 1024 ** 3, free)
+            self.gigabytes = free
+            logging.info("Resizing to %s GB", self.gigabytes)
+            sys.stdout.flush()
```
ldoktor: this shouldn't be necessary as logging handles per-line flushes.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if self.memory_mb > self.chunk_mb:
+            self.error("Too much RAM (%dMB) for this test to work" %
+                       self.memory_mb)
+
+        chunks = (1024 * self.gigabytes) / self.chunk_mb
```
ldoktor, line 130: the `self.chunk_mb` should be adjusted in case this assess to 0 (when the `self.gigabytes` is greater than `self.chunk_mb`).

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def one_disk_chunk(self, disks, chunk):
+        """
+        Tests one part of the disk by spawning a disktest instance.
+        :param disks: Directory (usually a mountpoint).
+        :param chunk: Portion of the disk used.
+        """
+        logging.info("Testing %d MB files on %s in %d MB memory, chunk %s",
+                     self.chunk_mb, disks, self.memory_mb, chunk)
+        logfile = self.outputdir + '/log.txt'
```
ldoktor: please use `os.path.join`

io/disk/disktest.py
ldoktor: Btw this actually does nothing as you create list of pids, but never add the pids to it. As you're creating just one pid at a time, just remove the pids and wait for each process. (again, please use proc.wait(), rather than os.waitpid).

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+        logging.info("Total of disk chunks that will be used: %s", chunks)
+        for i in range(chunks):
+            pids = []
+            pid = self.one_disk_chunk(self.disks, i)
+            errors = []
+            for pid in pids:
+                (junk, retval) = os.waitpid(pid, 0)
+                if retval != 0:
+                    errors.append(retval)
+            if errors:
```
ldoktor: Hmm is this suppose to finish early? If so than the handling could be simply to run it and let the process fail. If on the other hand you want to run all chunks and report all failures at once, then please initialize the `errors` before the for loop and add this branch after the for loop.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+                self.error("Errors from children: %s" % errors)
```
ldoktor: Hmm, isn't this actually the expected failure? I think this should be `self.fail` as the test run probably fine, but the disk reported errors.

(error => unexpected exception occurred; fail => the thing under stressing failed)

io/disk/disktest.py
ldoktor: Oh, it could be the the `self.workdir` as the disktest uses the directory and creates a `testfile` in it. Anyway please use `self.workdir` as default.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+            logging.info("Resizing to %s GB", self.gigabytes)
+            sys.stdout.flush()
+
+        if self.memory_mb > self.chunk_mb:
+            self.error("Too much RAM (%dMB) for this test to work" %
+                       self.memory_mb)
+
+        chunks = (1024 * self.gigabytes) / self.chunk_mb
+        logging.info("Total of disk chunks that will be used: %s", chunks)
+        for i in range(chunks):
```
ldoktor: nitpick: When you don't need the `range` value, but you only iterate through it, you should use `xrange` instead, which consumes less memory as it's iterator vs. list.

io/disk/disktest.py
```diff
@@ -0,0 +1,115 @@
...
+            self.gigabytes = free
+            logging.info("Resizing to %s GB", self.gigabytes)
+            sys.stdout.flush()
+
+        if self.memory_mb > self.chunk_mb:
+            self.error("Too much RAM (%dMB) for this test to work" %
+                       self.memory_mb)
+
+        chunks = (1024 * self.gigabytes) / self.chunk_mb
+        logging.info("Total of disk chunks that will be used: %s", chunks)
```
ldoktor: @Nosheen08 up to this point it's still the initialization. So please move the above to the `setUp` part and use `self.skip` to mark test as `SKIP` when it does not match the specification (instead of `error`).

ldoktor:
Hello @Nosheen08, I'm sorry I didn't noticed you pushed a new version. Let me take a look now...

ldoktor:
OK I spent quite a time on this and it mixes the old one-disk version with the new multi-disks version, resulting in many weird constructs. So before pointing out all the details, I'd like to get an input on what is your plan and what do you expect from this test.

The original test supported multiple `disks` argument to synchronously check the chunks in parallel (first chunk on all disks, checks for failures, then second chunk on all disks, ...).

This version still asks for `disks`, but in fact only supports one disk. Then it runs chunk by chunk check and reports list of failed returncodes without any useful info like which chunk actually failed.

So I understand you might not need the multiple disks support, then the test should be simplified significantly. On the other hand as the test already supports it, it'd be easier and nicer to add the test as is and only replace the autotest to avocado API.

Additionally. I looked at the history of the file and the `/8` when calculating the memory is IMO a bug introduced while debugging the test by @jzupka in 76ff2606e9cbf6d5b616801991b461f475dd80aa.

Regarding the license, only the original author should be listed: `Martin J. Bligh <mbligh@google.com>` 2007, commit 3bba4eec38d9600f3d41a85dd1937311dba99f48. The other guys made just minor changes and they did not claim the contribution.

Nosheen08:
Hi @ldoktor the disktest is taking one disk at a time, and tests on chunks on that disk.I am using the yaml file to get all disks, since there is no need to run on all disks at a time.For failure on chunks, I will provide the chunk on which test fails. Will this be fine ?

Nosheen08:
I could not understand your comment about bug introduced by @jzupka. Could you please eloborate it So it would be easy to make it correct

io/disk/disktest.py
```diff
@@ -0,0 +1,135 @@
...
+        chunks = (1024 * self.gigabytes) / self.chunk_mb
+        self.log.info("Total of disk chunks that will be used: %s", chunks)
+        errors = []
+        for i in xrange(chunks):
+            proc = self.one_disk_chunk(self.disks, i)
+            retval = proc.wait()
+            if retval != 0:
+                errors.append(retval)
+        if errors:
+            self.fail("Errors from children: %s" % errors)
```
ldoktor, line 139: Hello @Nosheen08, let's put the original test here:

``` python
        os.chdir(self.disk_srcdir)
        if chunk_mb is None:
            chunk_mb = utils_memory.memtotal() / 1024 / 8
        if disks is None:
            disks = [self.tmpdir]
        if gigabytes is None:
            free = 100  # cap it at 100GB by default
            for disk in disks:
                free = min(utils.freespace(disk) / 1024 ** 3, free)
            gigabytes = free
            logging.info("Resizing to %s GB", gigabytes)
            sys.stdout.flush()

        self.chunk_mb = chunk_mb
        self.memory_mb = utils_memory.memtotal() / 1024 / 8
        if self.memory_mb > chunk_mb:
            raise error.TestError("Too much RAM (%dMB) for this test to work" %
                                  self.memory_mb)

        chunks = (1024 * gigabytes) / chunk_mb

        logging.info("Total of disk chunks that will be used: %s", chunks)
        for i in range(chunks):
            pids = []
            for disk in disks:
                pid = self.test_one_disk_chunk(disk, i)
                pids.append(pid)
            errors = []
            for pid in pids:
                (junk, retval) = os.waitpid(pid, 0)
                if (retval != 0):
                    errors.append(retval)
            if errors:
                raise error.TestError("Errors from children: %s" % errors)
```

compare to your version, there is one significant difference. The original version supports list of `disks`, for example `["/dev/sdb", "/dev/sdc", "/tmp/foo", "/tmp/bar"]`. Because of this the code is more complex to use the smallest size to fit in all disks, it supports running the `one_disk_chunk` in parallel and so on.

Your version keeps some of the complexity, but only allows single disk. In some cases that results in unnecessary code, which is acceptable if you intend to extend it in the future. But the main testing loop which spawns the `one_disk_chunk` is actually broken, because:
1. `one_disk_chunk` itself waits for command completion (I guess it's remnant from your debugging, because the main loop also waits for proc to finish).
2. You don't fail early, but append the `one_disk_chunk` test return value to errors and report them after all chunks are tested. That's a change from the original test, which failed with first broken chunk, not necessarily a mistake if it's really intended. The problem is, that the list of errors says nothing, because you don't know which chunk failed.

So I'd suggest porting the test as was originally implemented (with the support for multiple disks) and only adjust the bits related to avocado like `process`, exceptions and style. Alternatively you can say you're only interested in simple version, then feel free to implement it, but properly with reasonable error messages and without unnecessary code.

ldoktor:
Regarding the @jzupka's potential bug: I wondered why the `chunk_mb` is evaluated as `memory.memtotal() / 1024 / 8`. So I went back in history and noticed, that it always was just `memory.memtotal() / 1024` until the commit 76ff2606e9cbf6d5b616801991b461f475dd80aa:

```
commit 76ff2606e9cbf6d5b616801991b461f475dd80aa
Author: Jiří Župka <jzupka@redhat.com>
Date:   Thu Nov 15 13:51:25 2012 +0100

    client: Move disktest to test dependencies.

    Disktest is very useful and could be used in whole autotest.

    Signed-off-by: Jiří Župka <jzupka@redhat.com>
```

Which does not say anything about that value. So I believe he used the `/8` to speedup the execution on his machine while developing the modification and then forgot to remove it. Until now nobody noticed that, so let's fix it here. (use `self.chunk_mb = memory.memtotal() / 1024`)

Of course I could be wrong, so if you know why the `/8` then please let me know.

Nosheen08:
Hi @ldoktor this test takes one disk at one time and variants can be used for multiple disks. I have created new pull request with the changes you suggested and documented the needed information. Please review it and let me know if more changes needed. Now I ll close this request.

PR #29, opened 2016-05-23
Ported Tiobench test from autotest to avocado.

rajashreer7:
Added tiobench test for avocado.

### DIFF
--- a/io/disk/tiobench.py
+++ b/io/disk/tiobench.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM.
+# Author: Rajashree Rajendran<rajashr7@linux.vnet.ibm.com>
+
+# Based on code by
+#   Author: Lucas Meneghel Rodrigues <lmr@redhat.com>
+#   Author: Xu Tian <xutian@redhat.com>
+#   Author: Jeff Moyer <jmoyer@redhat.com>
+#   Author: Dmitry Monakhov <dmonakhov@openvz.org>
+#   Copyright: 2008 - 2015 Red Hat, Inc.
+#   Copyright: 2013 Red Hat, Inc.
+#   Copyright: 2010 Red Hat, Inc.
+#   Copyright: 2012 <dmonakhov@openvz.org>
+#   https://github.com/autotest/autotest-client-tests/tree/master/tiobench
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import build
+from avocado.utils import git
+from avocado.utils import genio
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Tiobench(Test):
+
+    """
+    Avocado module for tiobench.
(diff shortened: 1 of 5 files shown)


io/disk/tiobench.py
```diff
@@ -0,0 +1,96 @@
...
+        iterations = self.params.get('iterations', default=2)
+        results = process.system_output('./tiobench.pl --target {} --block={}'
+                                        ' --threads={} --size={} --numruns={}'
+                                        .format(target, blocks, threads,
+                                                size, num_runs))
+        self.log.info(results)
+        for iteration in range(1, iterations + 1):
+            results_path = os.path.join(self.outputdir,
+                                        'raw_output_%s' % iteration)
+
```
sathnaga: @rajashreer7 Iterations does not have a meaning here until we make it explicitly handled.
Have done a quick if python unit test supports it, could not get anything of that sort, It use to be very well handled in autotest.
one of the way to address it, is there any better way to handle iterations from framework itself? @ldoktor @clebergnu 

```
def test(self):
    iterations = self.params.get('iterations', default=2)
    for iteration in range(1, iterations + 1):
        target = self.params.get('target', default='/mnt')
        blocks = self.params.get('blocks', default=4096)
        threads = self.params.get('threads', default=10)
        size = self.params.get('size', default=1024)
        num_runs = self.params.get('numruns', default=2)
        self.log.info(iterations)
        results = process.system_output('./tiobench.pl --target {} --block={}'
                                        ' --threads={} --size={} --numruns={}'
                                        .format(target, blocks, threads,
                                                size, num_runs))
        self.log.info(results)
        results_path = os.path.join(self.outputdir,
                                        'raw_output_%s' % iteration)
        genio.write_file(results_path, results)
```

ldoktor: Basically the multiplexer is way to run combinations. Right now there is no way to just provide a number of iterations, so you'd have to create the variants similarly to `examples/tests/doublefree_nasty.py.data/iterations.yaml`. But we plan to extend the multiplexer, currently it's about the internal filters, but then we can focus on specifying arguments by lists, or even range as those could be useful.

io/disk/tiobench.py.data/tiobench.yaml
```diff
@@ -0,0 +1,33 @@
...
+        default: 
+            size: 1024
+     
+    runs: !mux
+        default: 
+            numruns: 2
+        minimal:
+            numruns: 1
+  
+    iteration: !mux
```
ldoktor, line 24: To use native test iterations you could replace this with:

``` yaml
    iteration: !mux
        default:
        quick:
            1:
            2:
```

And remove the iterations handling completely from your test.

ldoktor, line 24: PS: In the future I can imagine a new tag to produce those automatically, eg:

``` yaml
iterations:
    quick:
        !range : 2
```

(that would be actually very simple change in avocado.core.tree, like 3-4 lines...)

rajashreer7, line 24: So, After this change, will this be accessible in a variable like `self.iteration` for usage in code?
And for now if we run it with this,

```
    iteration: !mux
        default:
        quick:
            1:
            2:
```

Is there a way to get the iteration number?

ldoktor, line 24: Nope. The test is a test, it knows nothing about whether it's run as first or second or as the only one. It should simply perform the test. Then in results you see that this test was executed twice with following params and you can query for test files. I see you're storing a results, there is a `self.whiteboard` available for such operations, which is queryable from results so that would be actually a good place for it. Then you could ask for whiteboard of the first and the second test and compare them (you could still do it without the whiteboard, it'd just be more convenient this way as whiteboard is part of the json results and is queryable from database).

ldoktor, line 24: Would this work for you? I mean if you want to evaluate the differences inside the test, then you have to create iterations. In case you just want to store the results and eventually compare them later, the multiplexer is the preferred solution.

PR #31, opened 2016-05-26
Added fs_mark tests.

kalshett:
Ported autotest fs_mark test case to avocado-misc-tests under io/disk.

### DIFF
--- a/io/disk/fs_mark.py.data/fs_mark.yaml
+++ b/io/disk/fs_mark.py.data/fs_mark.yaml
@@ -1,0 +1,23 @@
+# fs_mark options:
+#  -n 10000000: The number of files to be tested per thread (more on that later)
+#  -s 400; Each file will be 400KB
+#  -L 1: Loop the testing once (fs_mark testing)
+#  -S 0: Issue no sync() or fsync() during the creation of the file system. 
+#  -D 10000: There are 10,000 subdirectories under the main root directory
+#  -d /mnt: The root directory for testing; for this particular test, we are using only 1  #   root directory
+#   -N 1000: 1,000 files are allocated per directory
+#   -t 10: Use 10 threads for building the file system
+#   -k: Keep the file system after testing
+#   -N 10,000: Allocate 10,000 files per directory
+#   SAMPLE RUN
+#   A typical run of the program would look like this:
+# ./fs_mark -d /mnt -s 10240 -n 1000
+
+setup:
+    num_files:
+        num_files: 1000
+    size:
+        size: 10240
+    dir:
+        dir: /var/tmp
+
--- a/io/disk/fs_mark.py
+++ b/io/disk/fs_mark.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (C) 2003-2004 EMC Corporation
+#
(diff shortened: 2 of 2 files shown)


apahim:
It's failing here with permission denied:

```
Running './fs_mark -d /mnt -s 1000 -n 1024'
[stdout]
[stdout] #  ./fs_mark  -d  /mnt  -s  1000  -n  1024
[stdout] #      Version 3.3, 1 thread(s) starting at Thu May 26 15:52:46 2016
[stdout] #      Sync method: INBAND FSYNC: fsync() per file in write loop.
[stdout] #      Directories:  no subdirectories used
[stdout] #      File names: 40 bytes long, (16 initial bytes of time stamp with 24 random bytes at end of name)
[stdout] #      Files info: size 1000 bytes, written with an IO size of 16384 bytes per write
[stdout] #      App overhead is time in microseconds spent in the test not doing file writing related system calls.
[stdout]
[stdout] FSUse%        Count         Size    Files/sec     App Overhead
[stderr] Error in creat: Permission denied
[stderr] fopen failed to open: fs_log.txt.6980

Reproduced traceback from: /usr/lib/python2.7/site-packages/avocado-35.0-py2.7.egg/avocado/core/test.py:435
Traceback (most recent call last):
  File "/home/apahim/src/avocado-misc-tests/io/disk/fs_mark.py", line 64, in test
    process.run(cmd)
  File "/usr/lib/python2.7/site-packages/avocado-35.0-py2.7.egg/avocado/utils/process.py", line 1031, in run
    raise CmdError(cmd, sp.result)
CmdError: Command './fs_mark -d /mnt -s 1000 -n 1024' failed (rc=1)

Traceback (most recent call last):

  File "/usr/lib/python2.7/site-packages/avocado-35.0-py2.7.egg/avocado/core/test.py", line 490, in _run_avocado
    raise test_exception

CmdError: Command './fs_mark -d /mnt -s 1000 -n 1024' failed (rc=1)

ERROR 1-io/disk/fs_mark.py:fs_mark.test -> CmdError: Command './fs_mark -d /mnt -s 1000 -n 1024' failed (rc=1)
```

I guess you could use a distinct default dir?

io/disk/fs_mark.py.data/fs_mark.yaml
```diff
@@ -0,0 +1,23 @@
...
+#   A typical run of the program would look like this:
+# ./fs_mark -d /mnt -s 10240 -n 1000
+
+setup:
+    num_files:
+        num_files: 1000 
+    size:
+        size: 10240
+    dir:
+    	dir: /mnt
```
apahim: Please replace the tab by whitespaces.

kalshett: sure, I will make changes. tks.

kalshett: did you ran on x86 as non-root user?
also wanted to know which default dir would you suggest?

apahim: Yes, Fedora 23 Kernel 4.4.9-300.fc23.x86_64, non-root user, selinux enabled. If you will require root for this test, you can use in `setUp()` something like:

```
if os.geteuid() != 0:
    self.error(msg)
```

But that does not seem to be the case. I'd suggest any dir that does not make the test to fail by default... `/var/tmp` looks like a good candidate, what do you think?

kalshett: sounds good. It has passed for /var/tmp as non-root user. however later we see it needs permission for /proc, see below error:
      "Not logging /proc/slabinfo (lack of permissions)"
so looks like these kind of tests need to be run as root?  So we might have to do geteuid changes.

apahim:
No, this ("Not logging /proc/slabinfo (lack of permissions)") is not on your test. This is coming from sysinfo. You can just ignore it.

io/disk/fs_mark.py.data/fs_mark.yaml
apahim: No, this ("Not logging /proc/slabinfo (lack of permissions)") is not on your test. This is coming from sysinfo. You can just ignore it.

apahim:
Looks good. Thank you.

PR #32, opened 2016-05-27
Adding ethtool tests for Avocado

narasimhan-v:

### DIFF
--- a/io/net/ethtool/ethtool_tests.sh
+++ b/io/net/ethtool/ethtool_tests.sh
@@ -1,0 +1,63 @@
+#!/bin/bash -e
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+# Tests the network driver and interface with 'ethtool' command.
+# 12 different options of ethtool command are tested for each
+# of the interfaces specified in 'config' file
+
+PATH=$(avocado "exec-path"):$PATH
+
+# Install dependencies
+if [[ `python -c 'from avocado.utils.software_manager import SoftwareManager; \
+    print SoftwareManager().install("ethtool")'` == 'False' ]]
+then
+    avocado_debug 'Ethtool not installed'
+    exit
+fi
+
+CONFIG_FILE="$AVOCADO_TEST_DATADIR"/config
+interfaces=$(cat $CONFIG_FILE | grep -w "Interfaces" | awk -F'=' '{print $(NF)}')
+echo $interfaces
+
+for interface in $interfaces; do
+    echo $interface
+    ifup $interface
+    for command in ifdown ifup; do
+        sleep 10
+        $command $interface
+        sleep 10
+        echo "Ethool, $command:"
+        output_file=`mktemp`
+        echo $output_file
+        ethtool ${arg} ${interface} | tee $output_file
+        error=$?
+
+        if [ $error == 0 ]; then
+            echo "PASS"
+        elif grep "Operation not supported" $output_file; then
+            echo "PASS (Not supported return:$error)"
+        elif grep "no stats available" $output_file; then
+            echo "PASS (No stats available return:$error)"
+        elif [[ "${arg}" = "-t" && "$command" = "ifdown" ]]; then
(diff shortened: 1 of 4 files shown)


io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,63 @@
...
+
+PATH=$(avocado "exec-path"):$PATH
+
+# Install dependencies
+if [[ `python -c 'from avocado.utils.software_manager import SoftwareManager; \
+    print SoftwareManager().install("ethtool")'` == 'False' ]]
+then
+    avocado_debug 'Ethtool not installed'
+    exit
+fi
```
apahim, line 29: ok

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,63 @@
...
+
+CONFIG_FILE="$AVOCADO_TEST_DATADIR"/config
+interfaces=$(cat $CONFIG_FILE | grep -w "Interfaces" | awk -F'=' '{print $(NF)}')
```
apahim, line 32: Here you' re trusting IFS will have whitespace. Ok, I never had a problem with this.

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,63 @@
...
+echo $interfaces
```
apahim, line 33: I'd include some extra info in this echo. With only one interface in config file, this will echo the same information as in line 36 below. Something like `Device list: eth0 eth1`. And in line 36, `Testing device eth0`...

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,63 @@
...
+
+for interface in $interfaces; do
```
apahim, line 35: Why not use multiplex for this? You could save a good number of code lines putting the interfaces names in multiplex files and letting avocado interact over them. This would avoid the `for` loop here and also the config file parsing in line 32 (config file would not exist at all).

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,63 @@
...
+        if [ $error == 0 ]; then
+            echo "PASS"
+        elif grep "Operation not supported" $output_file; then
+            echo "PASS (Not supported return:$error)"
+        elif grep "no stats available" $output_file; then
+            echo "PASS (No stats available return:$error)"
+        elif [[ "${arg}" = "-t" && "$command" = "ifdown" ]]; then
+            echo "PASS (ethtool -t may fail when iface is down)"
+        else
+            echo "FAIL (return: $error)"
```
apahim, line 57: I think you should also set the avocado test to exit with `FAIL` instead of `PASS` if something goes bad with your inner  tests. You can have a variable called let say `any_error` and set it to true when some test enter in this else. Then, after the `fi` in line 58, you can test the `any_error` and exit script with a non-zero exit code, so avocado can show `FAIL` for the user, meaning some inner test failed.

io/net/ethtool/ethtool_tests.sh.data/README.txt
```diff
@@ -0,0 +1,14 @@
+Tests the network driver and interface with 'ethtool' command.
+12 different options of ethtool command are tested for each
```
apahim, line 2: I don't see 12 tests in the test script. The number of tests depends on the multiplex file. Please clarify the multiplex usage instead.

io/net/ethtool/ethtool_tests.sh.data/README.txt
```diff
@@ -0,0 +1,14 @@
...
+of the interfaces specified in 'config' file
```
apahim, line 3: If you' re going to use the multiplex file for the devices, remember to drop this line.

io/net/ethtool/ethtool_tests.sh.data/README.txt
```diff
@@ -0,0 +1,14 @@
...
+
+Requirements:
+-------------
+For all specified interfaces, configuration file for that interface needs
+to be updated, so that 'ifup/ifdown <interface>' configures the interface.
+
+Input Needed (in 'config' file):
```
apahim, line 10: ...and this section. Clarify then how to use the multiplex file for the devices.

apahim:
@narasimhan-v, thank you for this PR. I have some comments for your appreciation.

io/net/ethtool/ethtool_tests.sh
narasimhan-v, line 33: Thanks. Will implement this.

io/net/ethtool/ethtool_tests.sh
narasimhan-v, line 35: Thanks. Will implement this.

io/net/ethtool/ethtool_tests.sh
narasimhan-v, line 57: I have handled most of the errors due to unsupported configurations.
I have also exited avocado when the individual commands fail.

narasimhan-v:
Created a new, with proposed changes implemented.
PR #37
Closing this PR.

PR #33, opened 2016-05-27
Adding scsi remove and add test for Avocado

narasimhan-v:


apahim:
Ok, this is similar to the ethtool PR, so the same suggestion applies  here (use multiplex for the pci_devices interactions). I will review this better after we agree on the approach for the interactions.

narasimhan-v:
@apahim,  Have incorporated the changes suggested for ethtool PR.
PR #50

PR #35, opened 2016-05-27
Added service_check which turns ON/OFF services taken from .cfg file

santhosh93:
The Script parses the services list  from the respective distro section of services.cfg file and does the ON/OFF tests.

### DIFF
--- a/generic/service_check.py
+++ b/generic/service_check.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Santhosh G <santhog4@linux.vnet.ibm.com>
+#
+# Copyright: 2014 Red Hat Inc.
+# Besed on the Sample Idea from:
+# https://github.com/autotest/virt-test/blob/master/samples/service.py
+
+import ConfigParser
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils.service import SpecificServiceManager
+from avocado.utils import distro
+from avocado.utils.wait import wait_for
+
+
+class service_check(Test):
+
+    def test(self):
+        detected_distro = distro.detect()
+        parser = ConfigParser.ConfigParser()
+        config_file = self.datadir + '/services.cfg'
+        parser.read(config_file)
+        services_list = parser.get(detected_distro.name, 'services').split(',')
+        services_failed = []
+        runner = process.run
+
+        for service in services_list:
+            service_obj = SpecificServiceManager(service, runner)
+            self.log.info("Checking %s service" % service)
+            if service_obj.is_enabled() is False:
+                self.log.info("%s service Not Found !!!" % service)
+                services_failed.append(service)
+                continue
+            original_status = service_obj.status()
+            if original_status is True:
+                service_obj.stop()
+                if not wait_for(lambda: not service_obj.status(), 10):
+                    self.log.info("Fail to stop %s service" % service)
+                    services_failed.append(service)
+                    continue
+                service_obj.start()
+                wait_for(service_obj.status, 10)
+            else:
+                service_obj.start()
+                if not wait_for(service_obj.status, 10):
+                    self.log.info("Fail to start %s service" % service)
+                    services_failed.append(service)
+                    continue
+                service_obj.stop()
+                wait_for(lambda: not service_obj.status(), 10)
+            if not service_obj.status() is original_status:
+                self.log.info("Fail to restore original status of the %s"
+                              "service" % service)
+                services_failed.append(services)
+
+        if services_failed:
+            self.fail("List of services failed: %s" % services_failed)
+        else:
+            self.log.info("All Services Passed the ON/OFF test")
+
+if __name__ == "__main__":
+    main()
--- a/generic/service_check.py.data/services.cfg
+++ b/generic/service_check.py.data/services.cfg
@@ -1,0 +1,8 @@
+[Ubuntu]
+services = sshd,ufw,networking,libvirtd
+[redhat]
+services = sshd,firewalld,network,libvirtd
+[SuSE]
+services = sshd,SuSEfirewall2,network,libvirtd
+[fedora]
+services = sshd


generic/service_check.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class service_check(Test):
+
+    def test(self):
+        detected_distro = distro.detect()
+        parser = ConfigParser.ConfigParser()
+        config_file = self.datadir + '/services.cfg'
+        parser.read(config_file)
+        services_list = parser.get(detected_distro.name, 'services').split(',')
```
ldoktor, line 36: It might be friendlier to catch the `NoSectionError` exception and tell the user what happened (consider this a mild suggestion as the message is sort-of informative enough).

generic/service_check.py
```diff
@@ -0,0 +1,77 @@
...
+            service_obj = SpecificServiceManager(service, runner)
+            self.log.info("Checking %s service" % service)
+            if service_obj.is_enabled() is False:
+                self.log.info("%s service Not Found !!!" % service)
+                services_failed.append(service)
+                continue
+            original_status = service_obj.status()
+            if original_status is True:
+                service_obj.stop()
+                time.sleep(5)
```
ldoktor: Would you please use `avocado.utils.wait.wait_for` instead of hardcoded `sleep`? Also I'd consider 10s deadline as some services might take longer. The modification for the enabled service would look like this:

``` diff
diff --git a/generic/service_check.py b/generic/service_check.py
index 09a1034..94bfedd 100644
--- a/generic/service_check.py
+++ b/generic/service_check.py
@@ -23,6 +23,7 @@ from avocado import Test
 from avocado import main
 from avocado.utils.service import SpecificServiceManager
+from avocado.utils.wait import wait_for
 from avocado.utils import distro


@@ -47,12 +48,12 @@ class service_check(Test):
             original_status = service_obj.status()
             if original_status is True:
                 service_obj.stop()
-                time.sleep(5)
-                if service_obj.status() is not False:
+                if not wait_for(lambda: not service_obj.status(), 10):
                     self.log.info("Fail to stop %s service" % service)
                     services_failed.append(service)
                     continue
                 service_obj.start()
+                wait_for(service_obj.status, 10)
             else:
                 service_obj.start()
                 time.sleep(5)
@@ -62,7 +63,6 @@ class service_check(Test):
                     services_failed.append(service)
                     continue
                 service_obj.stop()
-            time.sleep(5)
             if not service_obj.status() is original_status:
                 self.log.info("Fail to restore original status of the %s"
                               "service" % service)
```

you can adjust the other branch accordingly.

generic/service_check.py
```diff
@@ -0,0 +1,77 @@
...
+                    continue
+                service_obj.stop()
+            time.sleep(5)
+            if not service_obj.status() is original_status:
+                self.log.info("Fail to restore original status of the %s"
+                              "service" % service)
+                services_failed.append(services)
+
+        if services_failed:
+            self.error("List of services failed: %s" % services_failed)
```
ldoktor: This should be `self.fail`:

ERROR -> unexpected generic failure
FAIL -> the item under testing failed

generic/service_check.py.data/services.cfg
```diff
@@ -0,0 +1,6 @@
+[Ubuntu]
+services = sshd,ufw,networking,libvirtd
+[redhat]
+services = sshd,firewalld,network,libvirtd
```
ldoktor, line 4: Would you mind adding:

```
[fedora]
services = sshd
```

santhosh93:
@ldoktor : Have incorporated the review comments...and pushed the changes.
NoSectionError exception is raised automatically if the section is not found.Still do you want me to raise an exception for that ?

generic/service_check.py
```diff
@@ -0,0 +1,74 @@
...
+# Copyright: 2014 Red Hat Inc.
+# Besed on the Sample Idea from:
+# https://github.com/autotest/virt-test/blob/master/samples/service.py
+
+import ConfigParser
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils.service import SpecificServiceManager
+from avocado.utils import distro
```
ldoktor, line 25: missing import `from avocado.utils.wait import wait_for`

generic/service_check.py
ldoktor, line 36: I think you can ignore this. Theoretically this failure should result in test `SKIP`, but other tests don't follow this either, so you can leave it as it is and people can adjust it in the future, should they need it.

generic/service_check.py
```diff
@@ -0,0 +1,74 @@
...
+                service_obj.start()
+                wait_for(service_obj.status, 10)
+            else:
+                service_obj.start()
+                if not wait_for(lambda: service_obj.status(), 10):
+                    self.log.info("Fail to start %s service" % service)
+                    services_failed.append(service)
+                    continue
+                service_obj.stop()
+                wait_for(lambda: not service_obj.status, 10)
```
ldoktor: Missing `()`, should be `wait_for(lambda: not service_obj.status(), 10)`

generic/service_check.py
```diff
@@ -0,0 +1,74 @@
...
+                service_obj.stop()
+                if not wait_for(lambda: not service_obj.status(), 10):
+                    self.log.info("Fail to stop %s service" % service)
+                    services_failed.append(service)
+                    continue
+                service_obj.start()
+                wait_for(service_obj.status, 10)
+            else:
+                service_obj.start()
+                if not wait_for(lambda: service_obj.status(), 10):
```
ldoktor: The lambda here is redundant, you can just use `wait_for(service_obj.status, 10)`

generic/service_check.py
```diff
@@ -0,0 +1,74 @@
...
+                    continue
+                service_obj.stop()
+                wait_for(lambda: not service_obj.status, 10)
+            if not service_obj.status() is original_status:
+                self.log.info("Fail to restore original status of the %s"
+                              "service" % service)
+                services_failed.append(services)
+
+        if services_failed:
+            self.fail("List of services failed: %s" % services_failed)
```
ldoktor, line 70: Thanks, this is better.

ldoktor:
Thank you for the changes, there are few typos making this test non-executable, but once you fix them I'm ready to merge it.

santhosh93:
@ldoktor : Hope this should work fine.

generic/service_check.py
```diff
@@ -0,0 +1,75 @@
...
+                continue
+            original_status = service_obj.status()
+            if original_status is True:
+                service_obj.stop()
+                if not wait_for(lambda: not service_obj.status(), 10):
+                    self.log.info("Fail to stop %s service" % service)
+                    services_failed.append(service)
+                    continue
+                service_obj.start()
+                wait_for(service_obj.status(), 10)
```
ldoktor: There is an extra `()` here. Basically `wait_for` accepts a function as the first argument and it calls the function until bool(returned_value) evaluates to True, then it returns the value (or timeout).

So without the lambda, you need to use `wait_for(service_obj.status, 10) => service_obj.status()`, which then is called by the `wait_for` in each iteration. When you use the lambda, you define a new function which is going to be called by `wait_for`, so you need to call the service_obj: `wait_for(lambda: not service_obj.status(), 10) => lambda() => not service_obj.status()`.

generic/service_check.py
```diff
@@ -0,0 +1,75 @@
...
+            else:
+                service_obj.start()
+                if not wait_for(service_obj.status(), 10):
```
ldoktor: the same problem here

santhosh93:
@ldoktor : I have made necessary changes and pushed. Can you please take a look at it ?

sathnaga:
ACKed

sathnaga:
@ldoktor Can you have this pull request merged if all comments addressed, Thanks in advance

ldoktor:
Thank you, merged.

PR #37, opened 2016-05-31
Adding ethtool tests for Avocado [v2]

narasimhan-v:

### DIFF
--- a/io/net/ethtool/ethtool_tests.sh
+++ b/io/net/ethtool/ethtool_tests.sh
@@ -1,0 +1,64 @@
+#!/bin/bash -e
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+# Tests the network driver and interface with 'ethtool' command.
+# Different parameters are specified in Parameters section of multiplexer file.
+# Interfaces are specified in Interfaces section of multiplexer file.
+
+PATH=$(avocado "exec-path"):$PATH
+
+# Install dependencies
+if [[ `python -c 'from avocado.utils.software_manager import SoftwareManager; \
+    print SoftwareManager().install("ethtool")'` == 'False' ]]
+then
+    avocado_debug 'Ethtool not installed'
+    exit
+fi
+
+echo "Interface: $interface"
+
+ip link show | grep $interface > /dev/null
+if [[ $? != 0 ]]; then
+    avocado_debug "$interface does not exist"
+    exit 1
+fi
+
+echo "Parameter: ${arg}"
+
+ifup $interface
+for command in ifdown ifup; do
+    sleep 10
+    $command $interface
+    sleep 10
+    echo "Ethool, $command:"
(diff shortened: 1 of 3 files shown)


io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,64 @@
...
+PATH=$(avocado "exec-path"):$PATH
+
+# Install dependencies
+if [[ `python -c 'from avocado.utils.software_manager import SoftwareManager; \
+    print SoftwareManager().install("ethtool")'` == 'False' ]]
+then
+    avocado_debug 'Ethtool not installed'
+    exit
+fi
+
```
apahim, line 30: Ok, so you agreed in use the multiplex to provide the interfaces you want to test.
When using the multiplex, a good practice is to not strictly depend on it. I mean, if no multiplex file is provided, your test should not fail badly.
Right now, if no multiplex file is provided, the test fails with `grep` error due to the missing `$interface`.
You can both set some defaults or test if the parameters from the multiplex are in place. For the interface, I'd suggest something like:

```
[[ -z $interface ]] && interface=eth1
```

The line above will set `eth1` as the interface to use if no multiplex file is provided.

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,64 @@
...
+echo "Interface: $interface"
+
+ip link show | grep $interface > /dev/null
+if [[ $? != 0 ]]; then
+    avocado_debug "$interface does not exist"
+    exit 1
+fi
```
apahim, line 37: We can avoid the `grep` here and also the `if` with:

```
ip link show dev $interface > /dev/null 2>&1 || {
    avocado_debug "$interface does not exist"
    exit 1
}
```

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,64 @@
...
+        avocado_debug "PASS"
+    elif grep "Operation not supported $output_file"; then
+        avocado_debug "PASS (Not supported return:$error)"
+    elif grep "no stats available $output_file"; then
+        avocado_debug "PASS (No stats available return:$error)"
+    elif [[ "${arg}" = "-t" && "$command" = "ifdown" ]]; then
+        avocado_debug "PASS (ethtool -t may fail when iface is down)"
+    else
+        avocado_debug "FAIL (return: $error)"
+        exit $error
```
apahim, line 61: Great, thank you.

io/net/ethtool/ethtool_tests.sh.data/README.txt
```diff
@@ -0,0 +1,13 @@
...
+
+Requirements:
+-------------
+For all specified interfaces, configuration file for that interface needs
+to be updated, so that 'ifup/ifdown <interface>' configures the interface.
+
+Input Needed (in multiplexer file):
+-----------------------------------
+Interfaces  -   Specify all the interfaces for which the test needs to be run.
+
```
apahim, line 13: Just noticed that you should mention somewhere that this test is expected to be executed as root.

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,64 @@
...
+echo "Parameter: ${arg}"
+
+ifup $interface
+for command in ifdown ifup; do
+    sleep 10
+    $command $interface
+    sleep 10
+    echo "Ethool, $command:"
+    output_file="$AVOCADO_TEST_OUTPUTDIR/output_$command"
+    ethtool ${arg} ${interface} $action_elapse | tee "$output_file"
```
apahim, line 48: For the `arg`, you can both set a deafult (for the case when multiplex is not provided) os just keep it as is right now, since missing `arg` will not break your test.

apahim:
@narasimhan-v, thank you for this v2. We are almost there. I have some additional comments.

narasimhan-v:
@apahim, Thanks for the comments.
I have made the changes in #49

PR #22, opened 2016-05-12
IOZone Test for avocado

harish-24:
Ported IOZone test from autotest-client-tests to avocado-misc-tests

PR #38, opened 2016-05-31
Porting IOZone Test for avocado [v2]

harish-24:

Porting IOZone test with changes in #22

### DIFF
--- a/io/disk/iozone.py
+++ b/io/disk/iozone.py
@@ -1,0 +1,576 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harish <harisrir@linux.vnet.ibm.com>
+
+# Based on code by:
+#   author: Ying Tao <yingtao@cn.ibm.com>
+#   author: Lucas Meneghel Rodrigues <lmr@redhat.com>
+#   copyright: 2006 IBM
+#   copyright: 2008 Red Hat, Inc.
+
+import os
+import re
+import json
+import logging
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import process
+from avocado.utils import build
+from avocado.utils import distro
+from avocado.utils import data_structures
+from avocado.utils import astring
+from avocado.utils.software_manager import SoftwareManager
+
+
+_LABELS = ['file_size', 'record_size', 'write', 'rewrite', 'read', 'reread',
+           'randread', 'randwrite', 'bkwdread', 'recordrewrite', 'strideread',
+           'fwrite', 'frewrite', 'fread', 'freread']
+
+
+class IOzoneAnalyzer(object):
+
+    """
+    Analyze an unprocessed IOzone file, and generate the following types of
+    report:
+
+    * Summary of throughput for all file and record sizes combined
+    * Summary of throughput for all file sizes
+    * Summary of throughput for all record sizes
+
+    If more than one file is provided to the analyzer object, a comparison
+    between the two runs is made, searching for regressions in performance.
+    """
+
+    def __init__(self, log, list_files, output_dir):
+        self.list_files = list_files
+        if not os.path.isdir(output_dir):
+            os.makedirs(output_dir)
+        self.output_dir = output_dir
+        self.log = log
+        self.log.info("Results will be stored in %s", output_dir)
+
+    @staticmethod
+    def average_performance(results, size=None):
+        """
+        Flattens a list containing performance results.
+        :param results: List of n lists containing data from performance runs.
+        :param size: Numerical value of a size (say, file_size) that was used
+                     to filter the original results list.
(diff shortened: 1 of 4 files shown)


io/disk/iozone.py
```diff
@@ -0,0 +1,577 @@
...
+        """
+        Parse an IOzone results file.
+
+        :param file: File object that will be parsed.
+        :return: Matrix containing IOzone results extracted from the file.
+        """
+        lines = []
+        for line in p_file.readlines():
+            fields = line.split()
+            if len(fields) != 15:
```
ldoktor, line 121: OK, this is going to fail in case of zero-division error as the value is `Error %s/%s`. I don't think we should adjust this code, though. Maybe we could improve the `compare_matrices` to produce splitable output, for example: `error_%s/%s` would work fine. @harish-24 what do you think? Should I send the fix to avocado, or do you have another suggestion?

io/disk/iozone.py
```diff
@@ -0,0 +1,577 @@
...
+
+        cmd = os.path.join(self.srcdir, 'src', 'current', 'iozone')
+        self.results = process.system_output('%s %s' % (cmd, args))
+        self.auto_mode = ("-a" in args)
+        results_path = os.path.join(self.outputdir,
+                                    'raw_output')
+        analysisdir = os.path.join(self.outputdir,
+                                   'analysis')
+        r_file = open(results_path, 'w')
+        try:
```
ldoktor: nitpick: python 2.6 supports `with` statement just for this purpose:

``` python
with open(results_path, 'w') as r_file:
    r_file.write(self.results)
```

But this works fine so take it as a suggestion for the future.

io/disk/iozone.py
```diff
@@ -0,0 +1,577 @@
...
+    If more than one file is provided to the analyzer object, a comparison
+    between the two runs is made, searching for regressions in performance.
+    """
+
+    def __init__(self, list_files, output_dir):
+        self.list_files = list_files
+        if not os.path.isdir(output_dir):
+            os.makedirs(output_dir)
+        self.output_dir = output_dir
+        logging.info("Results will be stored in %s", output_dir)
```
ldoktor: Oh, this whole test uses the `root` python logger, while tests should use the `avocado.test` logger instead. The recommended way is to use the log from `Test.log` as it contains additional features to detect `warn` usage.

If you don't want to pass the log from the test itself, you could also use `logging.getLogger("avocado.test")`, but passing the log from test would be preferred.

io/disk/iozone.py
```diff
@@ -0,0 +1,577 @@
...
+        output files for further processing (graph generation).
+
+        :param overall_results: 1x15 Matrix containing IOzone results for all
+                file sizes
+        :param record_size_results: nx15 Matrix containing IOzone results for
+                each record size tested.
+        :param file_size_results: nx15 Matrix containing file size results
+                for each file size tested.
+        """
+        logger = logging.getLogger()
```
ldoktor: The same here. Ideally you should pass the `self.log` from test and use it here as `logger`, or at least please use `getLogger("avocado.test")`. The difference is in the `debug.log` vs. `job.log`.

io/disk/iozone.py
```diff
@@ -0,0 +1,577 @@
...
+
+    Plots graphs based on the results of an IOzone run. Uses gnuplot to
+    generate the graphs.
+    """
+
+    def __init__(self, results_file, output_dir):
+        self.active = True
+        s_mg = SoftwareManager()
+        if not s_mg.check_installed("gnuplot") and not s_mg.install("gnuplot"):
+            logging.error("Command gnuplot not found, disabling graph "
```
ldoktor: and also here, basically you should not use `logging` directly, unless necessary.

ldoktor:
Hello @harish-24, you did a great job, it looks way better now and you even fixed some things I haven't complain about :smile:. The only problem I see is the direct use of `logging`, which produces different results for the overall `job.log` vs. test's `debug.log`. I'd suggest passing the `self.log` to the objects and using it there, but I could also accept simple `log = logging.getLogger("avocado.test")` and then using the `log`.

io/disk/iozone.py
lmr: Yep, this code was originally written to support py 2.4.

harish-24: Sure, will use this in future.

io/disk/iozone.py
harish-24, line 121: @ldoktor Improving compare_matrices and corresponding test in selftests would be good. You can send the fix to avocado.

io/disk/iozone.py
harish-24: @ldoktor  So, are you suggesting to pass self.log from Test class 

```
analysis = IOzoneAnalyzer(self.log, list_files=[results_path,
                                                      previous_results],
                                          output_dir=analysisdir)
```

and assigning self.log in `__init__()` of other classes?

ldoktor: yes, that would be the preferred solution.

harish-24:
Have done the changes. Thanks.

io/disk/iozone.py.data/README
```diff
@@ -0,0 +1,11 @@
+IOzone is a filesystem benchmark tool. The benchmark generates and measures
+a variety of file operations. Iozone has been ported to many machines and
+runs under many operating systems.
+
+Inputs Needed in yaml file:
+---------------------------
+args - Arguements with which iozone command is to be run.
+directory - Directory from which iozone test is executed.
+previous_run - Absolute path of raw_output file of any previously ran iozone
```
ldoktor, line 9: This should probably be `previous_results`, it's a minor change, I'll fix it during merge.

ldoktor:
Fixed the typo in README and merged. Thank you, @harish-24.

PR #39, opened 2016-06-01
Ported tiobench test from autotest to avocado [v2].

rajashreer7:
Added tiobench test for avocado with changes in #29 

Signed-off-by: Rajashree rajashr7@linux.vnet.ibm.com

### DIFF
--- a/io/disk/tiobench.py
+++ b/io/disk/tiobench.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM.
+# Author: Rajashree Rajendran<rajashr7@linux.vnet.ibm.com>
+
+# Based on code by
+#   Author: Lucas Meneghel Rodrigues <lmr@redhat.com>
+#   Author: Xu Tian <xutian@redhat.com>
+#   Author: Jeff Moyer <jmoyer@redhat.com>
+#   Author: Dmitry Monakhov <dmonakhov@openvz.org>
+#   Copyright: 2008 - 2015 Red Hat, Inc.
+#   Copyright: 2013 Red Hat, Inc.
+#   Copyright: 2010 Red Hat, Inc.
+#   Copyright: 2012 <dmonakhov@openvz.org>
+#   https://github.com/autotest/autotest-client-tests/tree/master/tiobench
+
+"""
+This program runs a multi-threaded I/O benchmark test
+to measure file system performance.
+"""
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import build
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Tiobench(Test):
+    """
+    Avocado module for tiobench.
+    """
+    def setUp(self):
+        """
+        Build tiobench.
+        Source:
+        https://github.com/mkuoppal/tiobench.git
+        """
+        s_mngr = SoftwareManager()
+        if not s_mngr.check_installed("gcc") and not s_mngr.install("gcc"):
+            self.error('Gcc is needed for the test to be run')
+        locations = ["https://github.com/mkuoppal/tiobench/archive/master.zip"]
+        tarball = self.fetch_asset("tiobench.zip", locations=locations)
+        archive.extract(tarball, self.srcdir)
+        os.chdir(os.path.join(self.srcdir, "tiobench-master"))
+        build.make(".")
+
+    def test(self):
+        """
+        Test execution with necessary arguments.
+        :params target: The directory in which to test.
+                        Defaults to ., the current directory.
+        :params blocks: The blocksize in Bytes to use. Defaults to 4096.
+        :params threads: The number of concurrent test threads.
+        :params size: The total size in MBytes of the files may use together.
+        :params num_runs: This number specifies over how many runs
+                          each test should be averaged.
+        """
+        target = self.params.get('target', default='/var/tmp')
+        blocks = self.params.get('blocks', default=4096)
+        threads = self.params.get('threads', default=10)
+        size = self.params.get('size', default=1024)
+        num_runs = self.params.get('numruns', default=2)
+        self.whiteboard = process.system_output('perl ./tiobench.pl '
+                                                '--target {} --block={} '
+                                                '--threads={} --size={} '
+                                                '--numruns={}'
+                                                .format(target, blocks,
+                                                        threads, size,
+                                                        num_runs))
+if __name__ == "__main__":
+    main()
--- a/io/disk/tiobench.py.data/tiobench.yaml
+++ b/io/disk/tiobench.py.data/tiobench.yaml
@@ -1,0 +1,25 @@
+setup:
+    location: !mux
+        default:
+            target: '/mnt'
+    block: !mux
+        first:
+            blocks: 4096
+        second:
+            blocks: 8192
+    thread: !mux
+        default:
+            threads: 10
+    blocksize: !mux
+        default:
+            size: 1024
+    runs: !mux
+        default:
+            numruns: 2
+    iteration: !mux
+        default: !mux
+            1:
+            2:
+            3:
+            4:
+            5:


io/disk/tiobench.py
```diff
@@ -0,0 +1,93 @@
...
+
+# Based on code by
+#   Author: Lucas Meneghel Rodrigues <lmr@redhat.com>
+#   Author: Xu Tian <xutian@redhat.com>
+#   Author: Jeff Moyer <jmoyer@redhat.com>
+#   Author: Dmitry Monakhov <dmonakhov@openvz.org>
+#   Copyright: 2008 - 2015 Red Hat, Inc.
+#   Copyright: 2013 Red Hat, Inc.
+#   Copyright: 2010 Red Hat, Inc.
+#   Copyright: 2012 <dmonakhov@openvz.org>
```
ldoktor, line 25: Might I ask you where you got these authors from? In autotest I see:
- client/tests/tiobench/control => `walkinair@cn.ibm.com`
- git log -- tests/tiobench => `Yao Fei Zhu <walkinair@cn.ibm.com>`

There are other contributors when I look in `git log`, but they never claimed authorship, so unless they do we don't have to include them. (GPL says we have to include the original author as well as everybody who claimed the authorship after a significant contribution).

io/disk/tiobench.py
```diff
@@ -0,0 +1,93 @@
...
+        :params num_runs: This number specifies over how many runs
+                          each test should be averaged.
+        :params iterations: Number of iterations the test needs to be run.
+        """
+        target = self.params.get('target', default='/var/tmp')
+        blocks = self.params.get('blocks', default=4096)
+        threads = self.params.get('threads', default=10)
+        size = self.params.get('size', default=1024)
+        num_runs = self.params.get('numruns', default=2)
+        self.whiteboard = process.system_output('./tiobench.pl --target {}'
```
ldoktor: There is a missing ` `after`--target {}`so the`--block` is part of it's value.

io/disk/tiobench.py
```diff
@@ -0,0 +1,93 @@
...
+                                                '--block={} --threads={}'
```
ldoktor: the same story on each line...

io/disk/tiobench.py
```diff
@@ -0,0 +1,93 @@
...
+                                                '--size={} --numruns={}'
+                                                .format(target, blocks,
+                                                        threads, size,
+                                                        num_runs))
+        self.log.info(self.whiteboard)
```
ldoktor: This is kind-of duplicate as the `process.system_output` already prints the output, so I'd suggest removing this one (alternatively using `verbose=False` in the `process.system_output`.

io/disk/tiobench.py
```diff
@@ -0,0 +1,93 @@
...
+        whiteboard_file = os.path.join(self.logdir, 'whiteboard')
+        genio.write_file(whiteboard_file, self.whiteboard)
```
ldoktor: This is not necessary, whiteboard is written by avocado automatically.

io/disk/tiobench.py.data/makefile.patch
```diff
@@ -0,0 +1,19 @@
+--- tiobench-0.3.3/Makefile.orig	2011-02-07 20:32:53.000000000 -0800
```
ldoktor: I don't see where this patch is applied.

io/disk/tiobench.py.data/tiotest.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/tiotest.c b/tiotest.c
```
ldoktor: The same story here, do you want to apply them, or is it just from autotest? I mean the compilation and execution works without it. (this one is actually just style fix, the Makefile fix could be serious if one wants to use custom CC)

ldoktor:
Hello @rajashreer7, thank you it's shaping really great. I noticed you're using git, how about using `asset` instead? The change is minimal and you only download zipped version and additionally it get's cached for the future. I tried that and this seems to work fine:

``` diff
diff --git a/io/disk/tiobench.py b/io/disk/tiobench.py
index 26bcf06..0fdfa36 100644
--- a/io/disk/tiobench.py
+++ b/io/disk/tiobench.py
@@ -29,6 +29,7 @@ import os

 from avocado import Test
 from avocado import main
 from avocado.utils import build
 from avocado.utils import git
@@ -57,10 +58,11 @@ class Tiobench(Test):
         s_mngr = SoftwareManager()
         if not s_mngr.check_installed("gcc") and not s_mngr.install("gcc"):
             self.error('Gcc is needed for the test to be run')
-        git.get_repo('https://github.com/mkuoppal/tiobench.git',
-                     destination_dir=self.srcdir)
-        os.chdir(self.srcdir)
-        build.make(self.srcdir)
+        locations = ["https://github.com/mkuoppal/tiobench/archive/master.zip"]
+        tarball = self.fetch_asset("tiobench.zip", locations=locations)
+        os.chdir(os.path.join(self.srcdir, "tiobench-master"))
+        build.make(".")

     def test(self):
         """
@@ -79,15 +81,13 @@ class Tiobench(Test):
         threads = self.params.get('threads', default=10)
         size = self.params.get('size', default=1024)
         num_runs = self.params.get('numruns', default=2)
-        self.whiteboard = process.system_output('./tiobench.pl --target {}'
-                                                '--block={} --threads={}'
+        self.whiteboard = process.system_output('perl ./tiobench.pl --target {} '
+                                                '--block={} --threads={} '
                                                 '--size={} --numruns={}'
                                                 .format(target, blocks,
                                                         threads, size,
                                                         num_runs))
-        self.log.info(self.whiteboard)
-        whiteboard_file = os.path.join(self.logdir, 'whiteboard')
-        genio.write_file(whiteboard_file, self.whiteboard)
+

 if __name__ == "__main__":
     main()
```

(you might need to polish it a bit, or feel free to ignore it if you have reasons for using git)

io/disk/tiobench.py.data/tiobench.yaml
```diff
@@ -0,0 +1,26 @@
...
+            blocks: 4096
+        second:
+            blocks: 8192
+
+    thread: !mux
+        default:
+            threads: 10
+
+    blocksize: !mux
+        default:
```
ldoktor: there are lots of empty characters in this file, please remove them. (for example at the end of this line)

io/disk/tiobench.py.data/README.txt
```diff
@@ -0,0 +1,13 @@
+This program runs  a multi-threaded I/O benchmark test to measure file system performance in four basic operations:
```
ldoktor: Do you insist on having this extra file here? I'd prefer putting this info as header docstring of the test itself, which is the first thing people see when opening the test. Additionally there are efforts in providing the test's docstring on dry-runs/inquire (not yet settled, but the idea is to have reliable source of information for each test without the need to opening the sources).

Anyway it's not a strong demand, perhaps a question on how are you using this info. Our QA uses the docstring to learn about the test.

io/disk/tiobench.py.data/tiobench.yaml
```diff
@@ -0,0 +1,26 @@
...
+        default: 
+            size: 1024
+     
+    runs: !mux
+        default: 
+            numruns: 2
+  
+    iteration: !mux
+        default:
+            iterations: 5
```
ldoktor: The `iterations` itself won't work. Either you have to list the iterations along with `!mux`:

``` yaml
    iterations: !mux
        default: !mux
            1:
            2:
            3:
            4:
            5:
```

or you need to implement the iterations inside the test.

I'm actually wondering why do you need iterations, I'd just increase the `numruns` if I wanted multiple results. What is the benefit in executing the test multiple times?

io/disk/tiobench.py
```diff
@@ -0,0 +1,93 @@
...
+        if not s_mngr.check_installed("gcc") and not s_mngr.install("gcc"):
+            self.error('Gcc is needed for the test to be run')
+        git.get_repo('https://github.com/mkuoppal/tiobench.git',
+                     destination_dir=self.srcdir)
+        os.chdir(self.srcdir)
+        build.make(self.srcdir)
+
+    def test(self):
+        """
+        Runs two iterations of test.
```
ldoktor: Really? I think it runs `tiobench` test once.

io/disk/tiobench.py
```diff
@@ -0,0 +1,93 @@
...
+        :params target: The directory in which to test.
+                        Defaults to ., the current directory.
+        :params blocks: The blocksize in Bytes to use. Defaults to 4096.
+        :params threads: The number of concurrent test threads.
+        :params size: The total size in MBytes of the files may use together.
+        :params num_runs: This number specifies over how many runs
+                          each test should be averaged.
+        :params iterations: Number of iterations the test needs to be run.
```
ldoktor: This param is not used.

io/disk/tiobench.py.data/tiobench.yaml
ldoktor: Btw you can verify the result by `avocado multiplex io/disk/tiobench.py.data/tiobench.yaml` (use `-h` to see options).

rajashreer7:
@ldoktor : Sure, will make those changes.

rajashreer7:
Hello @ldoktor, I have made those changes which were told. Please review it and let me know if there are any more changes to be made.

io/disk/tiobench.py
ldoktor, line 25: I still don't see the original author in this license.

io/disk/tiobench.py
```diff
@@ -0,0 +1,86 @@
...
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import build
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Tiobench(Test):
+    """
+    Avocado module for tiobench.
```
ldoktor, line 45: Avocado test for tiobench

io/disk/tiobench.py
```diff
@@ -0,0 +1,86 @@
...
+        Test execution with necessary arguments.
+        :params target: The directory in which to test.
+                        Defaults to ., the current directory.
+        :params blocks: The blocksize in Bytes to use. Defaults to 4096.
+        :params threads: The number of concurrent test threads.
+        :params size: The total size in MBytes of the files may use together.
+        :params num_runs: This number specifies over how many runs
+                          each test should be averaged.
+        """
+        target = self.params.get('target', default='/var/tmp')
```
ldoktor, line 73: Please use `self.workdir` to avoid clashes when running in parallel.

ldoktor, line 73: Actually I'd even change this to:

``` python
target = self.params.get('target') or self.workdir
```

To handle invalid input gracefully, what do you think? (eg. when one specifies `target: None` in the yaml file)

ldoktor:
Hello @rajashreer7 it's almost there. The license is still incorrect as the original author is someone else (you need to look at the original location, which is inside `autotest` and not `autotest-client-tests`. Only the first committer listed himself in the `autotest's` license, so we should reflect that (unless they claim their contribution).

Please take care of that, the tmp location and the docstring and we're ready to go.

io/disk/tiobench.py
rajashreer7, line 25: I couldn't find the author name Yao Fei Zhu walkinair@cn.ibm.com in git log and no copyright details are found. Can you tell me from where did you get the git log?

PR #43, opened 2016-06-02
Added ffsb test code from autotest

kalshett:
ported ffsb test code from autotest.

### DIFF
--- a/io/fs/README
+++ b/io/fs/README
@@ -1,0 +1,44 @@
+INTRODUCTION:
+=============
+This patch integrates the Flexible Filesystem Benchmark (FFSB) with
+the autotest. This integration enables the high-performace I/O load testing in the
+KVM Virtual machine environment.
+
+FFSB is a filesystem performance measurement tool.  It is a multi-threaded
+application (using pthreads), written entirely in C with cross-platform
+portability in mind.  It differs from other filesystem benchmarks in that
+the user may supply a profile to create custom workloads, while most other
+filesystem benchmarks use a fixed set of randomly generated workloads.
+
+More information about the FFSB can be got from reading the README in the
+FFSB source directory which can be accessed from this link:
+[http://sourceforge.net/projects/ffsb/]
+
+With this integration, it is now possible to test a variety of filesystems on
+the KVM guest for :
+
+(1) directed I/O with sequential/random read/write.
+(2) buffered I/O with sequential/random read/write.
+(3) use varying block alignment boundaries to measure filesystem behaviour.
+(4) use multithreaded workloads to stress the filesystem.
+(5) Exert weighted combination of I/O workloads to analyze the I/O performace
+    for a specific scenario.
+(6) Age filesystem according to a specified workload up to a specified limit.
+
+    Since the only interface used for the integration is a FFSB configuration file;
+Autotest will be able to run variety of I/O tests on guest as FFSB improves,
+with no or minimal code change in Autotest itself.
+
+USE:
+====
+To use the FFSB for Filesystem testing, two configuration files
+need to be modified -
+(1) Usual - tests.cfg file to activate the ffsb tests through KVM.
+(2) profile.cfg - where the workloads are specified.
+
+TODO:
+====
+* Add validations for max. number of threads according to number of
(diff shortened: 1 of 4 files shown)


io/fs/README
```diff
@@ -0,0 +1,44 @@
+INTRODUCTION:
+=============
+This patch integrates the Flexible Filesystem Benchmark (FFSB) with
+the autotest. This integration enables the high-performace I/O load testing in the
```
apahim, line 4: s/autotest/avocado

io/fs/README
```diff
@@ -0,0 +1,44 @@
...
+(2) buffered I/O with sequential/random read/write.
+(3) use varying block alignment boundaries to measure filesystem behaviour.
+(4) use multithreaded workloads to stress the filesystem.
+(5) Exert weighted combination of I/O workloads to analyze the I/O performace
+    for a specific scenario.
+(6) Age filesystem according to a specified workload up to a specified limit.
+
+    Since the only interface used for the integration is a FFSB configuration file;
+Autotest will be able to run variety of I/O tests on guest as FFSB improves,
+with no or minimal code change in Autotest itself.
```
apahim, line 30: s/Autotest/Avocado

io/fs/ffsb.py
```diff
@@ -0,0 +1,232 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 Red Hat, Inc.
+# Flexible Filesystem Benchmark (FFSB) is a cross-platform filesystem
+#  performance measurement tool
+# Ported to Avocado from Kalpana S Shetty(kalshett@linux.vnet.ibm.com)
+# source: https://github.com/autotest/autotest-client-tests/tree/master/ffsb
```
apahim, line 18: This can be improved. The original author seems to be here:
https://github.com/autotest/autotest-client-tests/blob/master/ffsb/control
Also, the copyright it not Red Hat for 2016, but IBM, since you are the one porting it. And before that, it's also IBM, since the original author's e-mail is from IBM as well. Here an example:
https://github.com/avocado-framework/avocado-misc-tests/blob/master/generic/error_cleanup.py

io/fs/ffsb.py
```diff
@@ -0,0 +1,232 @@
...
+            fw.write(line + '\n')
+        fr.close()
+        fw.close()
+
+    def setUp(self):
+        """
+        Build 'ffsb'.
+        """
+
+        self.update_config('profile.cfg.sample')
```
apahim, line 182: This `profile.cfg`  handling is quite confuse. I think you should:
- Rename `profile.cfg.sample`  to `profile.cfg.default`.
- Check if a custom `profile.cfg` is provided, using `profile.cfg.default` as default:

```
profile_cfg = self.params.get('profile_cfg', default='profile.cfg.default')
```
- After that, call `self.update_config(profile_cfg)` only once, here in `setUp()`.
- Drop the `self.update_config(cfg)` call from `test()`.

io/fs/ffsb.py
```diff
@@ -0,0 +1,232 @@
...
+        profile_src = os.path.join(self.datadir, filename)
+        profile_dst = os.path.join(os.path.dirname(self.datadir),
+                                   'profile.cfg')
+        shutil.copyfile(profile_src, profile_dst)
+
+    def test(self, cfg=None):
+        """
+        Runs a single iteration of the FFSB.
+        """
+        profile_cfg = self.params.get('profile_cfg', default='profile.cfg')
```
apahim, line 219: As long as we handle the profile.cfg file properly, we don't need to change it here. So we can always use `profile.cfg` to run ffsb, since `profile.cfg` will be a copy out of `profile.cfg.default` OR from the user custom profile.cfg file (provided through yaml file).

apahim:
Ping @kalshett. Closing this for PR for lack of activity. Please send a V2 with the requested improvements if you still have interest in have this merged.

kalshett:
sure, I will send the suggested comments incorporated this week.

PR #44, opened 2016-06-03
Adding Lvsetup Test for Avocado

harish-24:
Lvsetup Test from autotest to avocado with disk as input parameter.
Using [#1237](https://github.com/avocado-framework/avocado/pull/1237) from avocado as utility.

### DIFF
--- a/io/disk/lvsetup.py
+++ b/io/disk/lvsetup.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harish <harisrir@linux.vnet.ibm.com>
+#
+# Based on the code by:
+#
+# Copyright: 2012 Intra2net
+# Author: Plamen Dimitrov (plamen.dimitrov@intra2net.com)
+
+"""
+Test that automatically takes shapshots from existing logical volumes
+or creates them using a given policy.
+
+For details about the policy see below.
+"""
+
+from avocado import Test
+from avocado import main
+from avocado.utils import lv_utils
+
+
+class Lvsetup(Test):
+
+    """
+    Test class for creating logical volumes.
+    """
+
+    def test(self):
+        """
+        General logical volume setup.
+
+        The main part of the lvm setup checks whether the provided volume group
+        exists and if not, creates one from the ramdisk. It then creates a
+        logical volume if there is no logical volume, takes a snapshot from the
+        logical if there is logical volume but no snapshot, and merges with the
+        snapshot if both the snapshot and the logical volume are present.
+        """
+        # if no virtual group is defined create one based on ramdisk
+        disk = self.params.get('disk', default=None)
+        vg_name = self.params.get('vg_name', default='avocado_vg')
+        if disk:
+            vg_name = vg_name + '_' + disk.split('/')[-1]
+        lv_name = self.params.get('lv_name', default='avocado_lv')
+        lv_size = self.params.get('lv_size', default='1G')
+        lv_snapshot_name = self.params.get(
+            'lv_snapshot_name', default='avocado_sn')
+        lv_snapshot_size = self.params.get('lv_snapshot_size', default='1G')
+        ramdisk_vg_size = self.params.get('ramdisk_vg_size', default='10000')
+        ramdisk_basedir = self.params.get('ramdisk_basedir', default='/tmp')
+        ramdisk_sparse_filename = self.params.get(
+            'ramdisk_sparse_filename', default='virtual_hdd')
+        override_flag = self.params.get('override_flag', default=0)
+
+        if not lv_utils.vg_check(vg_name):
+            lv_utils.vg_ramdisk(disk, vg_name, ramdisk_vg_size,
+                                ramdisk_basedir,
+                                ramdisk_sparse_filename)
+        # if no snapshot is defined start fresh logical volume
+        if override_flag == 1 and lv_utils.lv_check(vg_name, lv_name):
+            lv_utils.lv_remove(vg_name, lv_name)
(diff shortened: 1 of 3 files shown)


io/disk/lvsetup.py
```diff
@@ -0,0 +1,99 @@
...
+            if (lv_utils.lv_check(vg_name, lv_snapshot_name) and
+                    lv_utils.lv_check(vg_name, lv_name)):
+                lv_utils.lv_revert(vg_name, lv_name, lv_snapshot_name)
+                lv_utils.lv_take_snapshot(vg_name, lv_name,
+                                          lv_snapshot_name,
+                                          lv_snapshot_size)
+
+            elif (lv_utils.lv_check(vg_name, lv_snapshot_name) and
+                  not lv_utils.lv_check(vg_name, lv_name)):
+                raise Exception("Snapshot origin not found")
```
ldoktor, line 87: Please use `self.fail; for expected failures. Also the message could be more user friendly...

io/disk/lvsetup.py.data/lvsetup.yaml
```diff
@@ -0,0 +1,31 @@
+setup:
```
ldoktor, line 1: Although this yaml file should work fine, I don't see it that important to create a category for each param. If there are no useful values then I don't think you even need to create one. And if you still want to add example, then it might be simplified. For example:

``` yaml
disk: None
vg_name: avocado_vg
lv_name: avocado_lv
size: !mux
    small:
        lv_size: "1M"
    big:
        lv_size: "1G"
...
```

In a real life I'd imagine even simpler yaml file:

``` yaml
scenario: !mux
    loopback:
        disk: None
        lv_size: "10M"
    real:
        disk: /dev/sde
        lv_size: None
```

rather than multiplexing all possible variants. Yes, one could do it, that doesn't mean he has to do it. Anyway I leave this on your consideration, it's meant to be an example, so I'm fine with either choice.

io/disk/lvsetup.py
```diff
@@ -0,0 +1,99 @@
...
+        vg_name = self.params.get('vg_name', default='avocado_vg')
+        if disk:
+            vg_name = vg_name + '_' + disk.split('/')[-1]
+        lv_name = self.params.get('lv_name', default='avocado_lv')
+        lv_size = self.params.get('lv_size', default='1G')
+        lv_snapshot_name = self.params.get(
+            'lv_snapshot_name', default='avocado_sn')
+        lv_snapshot_size = self.params.get('lv_snapshot_size', default='1G')
+        ramdisk_vg_size = self.params.get('ramdisk_vg_size', default='10000')
+        ramdisk_basedir = self.params.get('ramdisk_basedir', default='/tmp')
```
ldoktor, line 61: I'd strongly recommend `self.workdir` or `datadir.get_tmp_dir()`.

io/disk/lvsetup.py
harish-24, line 87: Sure.

io/disk/lvsetup.py
harish-24, line 61: Sure.

io/disk/lvsetup.py.data/lvsetup.yaml
harish-24, line 1: Just added all possible inputs for the test as a sample yaml file. I think the above simpler form will be fine. I will change it.

ldoktor:
OK, I was looking for the way it works and I'm sorry this is not acceptable at all. **The test execution must never alter the system state!** This test actually compounds of 3 tests based on the system status and 2 (incomplete) setup/cleanup functions.

So if you want to create a `snapshot` test, then you need
- `setUp` - where you check if the system status and fail if the volumes already exists (you might use flag to allow erasing, though). Additionally it might create the `vg_ramdisk`.
- `test` - one or more `test*` classes where you describe the scenario. Given the description, I'd suggest one test which would use the `ramdisk` or existing volume to `lv_create`, followed by `lv_take_snapshot`, then `lv_revert` and `lv_take_snapshot`.
- `tearDown` - which would go through created resources and it'd free them. That means if the volume was created by the test, remove it. If not, leave it intact. The same with all lvs, pvs, files, directories and everything else created in `setUp` and `test` classes. At any occasion the system should finish in the state it began.
- [optional] you can add other `test_*` methods to focus on different things. For example you can create `test_basic` which only creates and removes lv. Then `test_snapshot` which creates lv, snapshot and removes it. And then `test_revert_snapshot` which does the above and removes the snapshot in the end. Anyway I don't think it's necessarily to go into this much details. I'd start with `test_snapshot` and later other tests could be added focusing on other features (like `test_multiple_lvs`, `test_write_lv`, `test_multiple_pvs`, ...)

Regarding the failures, the `lv_utils` raises `LVException`, which results in `TestError`. Luckily for you, there is a support for `@avocado.fail_on(LVException)` decorator you should be using, which turns `LVException` into `TestFail`. To use it you just need to add it to `test`:

``` python
class LVSetup(Test):
    ...
    @avocado.fail_on(lv_utils.LVException)
    def test(self):
        ...
```

harish-24:
@ldoktor  Agreed. The idea was to port as it is from autotest and have the same flow. Will come up with a better flow as suggested.

ldoktor:
Yes, this test is in autotest, but I can't merge it without a proper cleanup. And while on the cleanup I think it's best to add the setup which should simply check the environment prior to testing to run the same test, otherwise one would have to run the test several times to check the features.

harish-24:
@ldoktor Is this work flow Ok?

```
setUp():
    # Check for VG
    # if present, self.skip or self.error stating existence
    # Check for LV
    # if present, self.skip or self.error stating existence
    # Check for snapshot
    # if present, self.skip or self.error stating existence
test_lv_create():
    # Create VG, LV
test_snapshot():
    # Create snapshot
    # lv_revert_with_snapshot
tearDown():
    # Remove PV, VG, LV and snapshots created by test.
```

ldoktor:
The problem with this is, that the `test_snapshot` would have to additionally create vg and lv. So for start I'd recommend creating just `test_snapshot` which would combine `lv_create` and `snapshot` tests.

As for the `tearDown` you need to check, whether the PV, VG, LV and snapshots were created by this test. If not, then you should not touch them. So I'd recommend something like:

``` python
def setUp(self):
    self.lv = None
    lv = "foo"
    if lv_exists(lv):
        self.skip("LV %s already exists...)
    self.lv = lv
    lv_create(lv)
...
def tearDown(self):
    if self.lv:
        try:
            lv_remove(self.lv)
...
```

This way when the LV already existed it stays intact. When it did not exist, it's attempted to be removed.

harish-24:
@ldoktor I have sent a PR with the new work-flow.

PR #46, opened 2016-06-03
Added dma_memtest for avocado

santhosh93:
Dma_memtest has been added.

### DIFF
--- a/memory/dma_memtest.py
+++ b/memory/dma_memtest.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python
+
+import os
+import shutil
+import re
+from avocado import Test
+from avocado.utils import process
+from avocado.utils import archive
+from avocado.core import data_dir
+
+
+class DmaMemtest(Test):
+    def setUp(self):
+        self.nfail = 0
+        tarball_base = self.params.get('tarball_base',
+                                       default='linux-2.6.18.tar.bz2')
+        parallel = self.params.get('parallel', default=True)
+        self.parallel = parallel
+        kernel_repo = 'http://www.kernel.org/pub/linux/kernel/v2.6'
+        tarball_url = os.path.join(kernel_repo, tarball_base)
+        tarball_md5 = '296a6d150d260144639c3664d127d174'
+        self.log.info('Downloading linux kernel tarball')
+        self.tarball = self.fetch_asset(tarball_url, asset_hash=tarball_md5,
+                                        algorithm='md5')
+        size_tarball = os.path.getsize(self.tarball) / 1024 / 1024
+
+        # Estimation of the tarball size after uncompression
+        compress_ratio = 5
+        est_size = size_tarball * compress_ratio
+        self.sim_cps = self.get_sim_cps(est_size)
+        self.log.info('Source file: %s', tarball_base)
+        self.log.info('Megabytes per copy: %s', size_tarball)
+        self.log.info('Compress ratio: %s', compress_ratio)
+        self.log.info('Estimated size after uncompression: %s', est_size)
+        self.log.info('Number of copies: %s', self.sim_cps)
+        self.log.info('Parallel: %s', parallel)
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


memory/dma_memtest.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
```
apahim, line 1: Please add licensing

memory/dma_memtest.py
```diff
@@ -0,0 +1,130 @@
...
+import os
+import shutil
+import re
+from avocado import Test
+from avocado.utils import process
+from avocado.utils import archive
+from avocado.core import data_dir
+
+
+class DmaMemtest(Test):
```
apahim, line 12: Please add doc-strings telling what this test is about.

memory/dma_memtest.py
```diff
@@ -0,0 +1,130 @@
...
+    def setUp(self):
+        self.nfail = 0
+        tarball_base = self.params.get('tarball_base',
+                                       default='linux-2.6.18.tar.bz2')
+        parallel = self.params.get('parallel', default=True)
+        self.parallel = parallel
+        kernel_repo = 'http://www.kernel.org/pub/linux/kernel/v2.6'
+        tarball_url = os.path.join(kernel_repo, tarball_base)
+        tarball_md5 = '296a6d150d260144639c3664d127d174'
```
apahim, line 21: Why are `kernel_repo` and `tarball_md5` hard coded while the `tarball_base` is configurable ? If one changes the `tarball_base` in yaml file, the download is doomed to fail.

apahim, line 21: Notice you don't have to have the yaml file. But if you do have it, don't let it break your test easily.

apahim:
Hi @santhosh93. Could you explain how this test works? I meant, how is DMA tested by making the OS swap and so on. You can include this as comments into the code or in commit message.
Thank you.

memory/dma_memtest.py.data/dma_memtest.yaml
```diff
@@ -0,0 +1,7 @@
+setup:
+    tarball_base: !mux
```
apahim, line 2: This is parametrization, so you can drop the `!mux` tag, unless you are thinking in run the test for more than one linux kernel.

memory/dma_memtest.py.data/dma_memtest.yaml
```diff
@@ -0,0 +1,7 @@
...
+        default:
+            tarball_base: linux-2.6.18.tar.bz2
+    parallel: !mux
```
apahim, line 5: ...same here.

memory/dma_memtest.py
```diff
@@ -0,0 +1,130 @@
...
+                self.log.error('Error comparing trees: %s', out_buf)
+
+        # Clean up for the next iteration
+        parallel_procs = []
+
+        self.log.info('Cleaning up')
+        for j in range(self.sim_cps):
+            tmp_dir = 'linux.%s' % j
+            shutil.rmtree(tmp_dir)
+        shutil.rmtree(base_dir)
```
apahim, line 124: Clean up routines are expected be in `tearDown()` so you can write it only once for all tests. You have only one test function here, but I'm pointing this just to keep it sane.

memory/dma_memtest.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def tearDown(self):
+        if self.nfail != 0:
+            self.fail('DMA memory test failed.')
+        else:
+            self.log.info('DMA memory test passed.')
```
apahim, line 130: This results are supposed to be inside the test. The `tearDown()` will be executed for every test you have. Again, you have only one test for while, but let's keep it sane here as well.

santhosh93:
@apahim: Created a v2 of this PR as https://github.com/avocado-framework/avocado-misc-tests/pull/56 with changes. So, closing this.

PR #54, opened 2016-06-09
Added rt-tests which runs signaltest as default

poojabs7:
rt-tests runs signaltest as default, other tests in rt-tests can be executed by adding entry in rt_test.yaml file.

### DIFF
--- a/perf/rt_tests.py
+++ b/perf/rt_tests.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import process
+from avocado.utils import git
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import distro
+
+
+class rt_tests(Test):
+
+    def setUp(self):
+        # Check for root permission
+        if os.geteuid() != 0:
+            exit("You need to have root privileges to run this script."
+                 "\nPlease try again, using 'sudo'. Exiting.")
+        # Check for basic utilities
+        sm = SoftwareManager()
+        detected_distro = distro.detect()
+        deps = ['gcc', 'git', 'make']
+        for package in deps:
+            if package == 'git' and detected_distro.name == "SuSE":
+                package = 'git-core'
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(package + ' is needed for the test to be run')
+        if detected_distro.name == "Ubuntu":
+            if not sm.check_installed("build-essential") and not \
+                     sm.install("build-essential"):
+                self.error('build-essential is needed for the test to be run')
+            if not sm.check_installed("libnuma-dev") and not \
+                    sm.install("libnuma-dev"):
+                self.error('libnuma-dev is needed for the test to be run')
+
+        if detected_distro.name == "redhat":
+            if not sm.check_installed("numactl-devel") and not \
+                                      sm.install("numactl-devel"):
+                self.error('numactl-devel is needed for the test to be run')
+        data_dir = os.path.abspath(self.datadir)
+        git.get_repo('git://git.kernel.org/pub/scm/utils/rt-tests/'
+                     'rt-tests.git', destination_dir=self.srcdir)
+        os.chdir(self.srcdir)
+        build.make(self.srcdir)
+
+    def test(self):
+        test_to_run = self.params.get('test_to_run', default='signaltest')
+        args = self.params.get('args', default='-t 10 -l 100000')
+        process.system('./' + test_to_run + ' ' + args)
+
+if __name__ == "__main__":
+    main()
--- a/perf/rt_tests.py.data/rt_tests.yaml
+++ b/perf/rt_tests.py.data/rt_tests.yaml
@@ -1,0 +1,8 @@
+setup:
+    general: !mux
+         default:
+             test_to_run: signaltest
+             args: '-t 10 -l 100000 '
+         cyclictest: !mux
+             test_to_run: cyclictest
+             args: '-t 10 -l 100000 '


perf/rt_tests.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
```
ldoktor, line 1: Would you please add license heading? If the test is inspired please do include the original author along with link to that file.

perf/rt_tests.py
```diff
@@ -0,0 +1,53 @@
...
+from avocado.utils import distro
+
+
+class rt_tests(Test):
+
+    def setUp(self):
+        # Check for root permission
+        if os.geteuid() != 0:
+            exit("You need to have root privileges to run this script."
+                 "\nPlease try again, using 'sudo'. Exiting.")
```
ldoktor, line 19: This is not necessary. We encourage people to run tests as users and the `process` utility has the option `sudo` to run commands as superuser (if configured).

perf/rt_tests.py
```diff
@@ -0,0 +1,53 @@
...
+        # Check for basic utilities
+        sm = SoftwareManager()
+        detected_distro = distro.detect()
+        deps = ['gcc', 'git', 'make']
```
ldoktor, line 23: how about:

``` python
deps = ["gcc", "make"]
if detected_distro.name == "Suse":
    deps.append("git-core")
else:
    deps.append("git")
if detected_distro.name == "Ubuntu":
    deps.append("build-essential")
    deps.append("libnuma-dev")
elif detected_distro.name == "redhat":
    deps.append("numactl-devel")
for package in deps:
    ...
```

perf/rt_tests.py
```diff
@@ -0,0 +1,53 @@
...
+                self.error('build-essential is needed for the test to be run')
+            if not sm.check_installed("libnuma-dev") and not \
+                    sm.install("libnuma-dev"):
+                self.error('libnuma-dev is needed for the test to be run')
+
+        if detected_distro.name == "redhat":
+            if not sm.check_installed("numactl-devel") and not \
+                                      sm.install("numactl-devel"):
+                self.error('numactl-devel is needed for the test to be run')
+        data_dir = os.path.abspath(self.datadir)
```
ldoktor, line 41: You probably wanted to use `self.workdir` or even `self.srcdir`.

perf/rt_tests.py
```diff
@@ -0,0 +1,53 @@
...
+        git.get_repo('git://git.kernel.org/pub/scm/utils/rt-tests/'
+                     'rt-tests.git', destination_dir=self.srcdir)
+        os.chdir(self.srcdir)
```
ldoktor, line 44: no need to change dir here, build does that for you.

perf/rt_tests.py
ldoktor, line 41: Actually the `data_dir` is not used, probably left over from development so please just remove it.

perf/rt_tests.py
```diff
@@ -0,0 +1,53 @@
...
+        build.make(self.srcdir)
+
+    def test(self):
+        test_to_run = self.params.get('test_to_run', default='signaltest')
+        args = self.params.get('args', default='-t 10 -l 100000')
+        process.system('./' + test_to_run + ' ' + args)
```
ldoktor, line 50: I'd prefer avoid using the `os.chdir` and use the abspath here. Also please use the sudo as I mentioned in setUp phase.

``` python
process.system("%s %s" % (os.path.join(self.srcdir, test_to_run), args),
               sudo=True)
```

ldoktor:
Hello @poojabs7, I'm sorry it took so long, we were quite cluttered with PR recently. Your test looks quite good, there are only few style modifications I spotted. Please take care of them and send a new pull request with a brief changelog and mention the link to this PR there so I get notified. (similar to https://github.com/avocado-framework/avocado/pull/1261 but the format is up to you...)

perf/rt_tests.py.data/rt_tests.yaml
```diff
@@ -0,0 +1,8 @@
+setup:
+    general: !mux
+         default:
+             test_to_run: signaltest
+             args: '-t 10 -l 100000 '
+         cyclictest: !mux
+             test_to_run: cyclictest
+             args: '-t 10 -l 100000 '
```
ldoktor, line 8: Could you please also add configuration for the other tests so we have all of them covered? There is an outstanding PR for pi_test https://github.com/avocado-framework/avocado-misc-tests/pull/71 so having all tests defined could help prevent the duplicity. I'd recommend something like:

```
pi_test:
    test_to_run: pi_stress
    args: "--duration=300"
```

for the `pi_test` and other suitable defaults for the rest of the tests.

perf/rt_tests.py
```diff
@@ -0,0 +1,53 @@
...
+                    sm.install("libnuma-dev"):
+                self.error('libnuma-dev is needed for the test to be run')
+
+        if detected_distro.name == "redhat":
+            if not sm.check_installed("numactl-devel") and not \
+                                      sm.install("numactl-devel"):
+                self.error('numactl-devel is needed for the test to be run')
+        data_dir = os.path.abspath(self.datadir)
+        git.get_repo('git://git.kernel.org/pub/scm/utils/rt-tests/'
+                     'rt-tests.git', destination_dir=self.srcdir)
```
ldoktor, line 43: there is a mirror allowing to download this file as `tgz`, I'd really prefer using `self.fetch_asset` rather than cloning the repo. The link is https://kernel.googlesource.com/pub/scm/utils/rt-tests/rt-tests/+archive/master.tar.gz

ldoktor, line 43: Btw my reason for trying to avoid the `git` is that:
1. git clones the whole history, which means more data
2. it's cloned per-test-variant, which means even more traffic

The `self.fetch_asset` caches the file so it's downloaded just once (it allows `expire` option to re-download outdated assets).

poojabs7:
Raised new PR https://github.com/avocado-framework/avocado-misc-tests/pull/93

PR #55, opened 2016-06-09
Check licensing on travis

apahim:
Let's try to avoid some interactions on the pull requests by checking
if some license strings are present in .py files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,4 @@ script:
     - inspekt lint
     - inspekt indent
     - inspekt style
+    - ./check_licensing.py
--- a/check_licensing.py
+++ b/check_licensing.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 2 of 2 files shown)


apahim:
@ldoktor could you take a look here? THe second commit is only for test purpose. I will remove it as soon as we agree on the first commit.

.travis.yml
```diff
@@ -20,3 +20,15 @@ script:
     - inspekt lint
     - inspekt indent
     - inspekt style
+    - |
+        err=false
+        for file in $(git diff --name-only $TRAVIS_COMMIT_RANGE -- *.py)
+        do
+            echo -n Checking license in file $file...
+            inspekt license --copyright "Copyright" --author "Author" $file || err=true
```
ldoktor: Nice, there is only one problem I see, inspekt only supports GPLv2+ license for now (even though it claims GPLv2). And we have GPLv2_ONLY code getting to avocado and others like LGPL might follow.

apahim: Yes, just noticed that as well. I will send a PR for inspekt, adding more license types.

apahim:
@ldoktor @clebergnu, what do you guys think about this version?

apahim:
@ldoktor @clebergnu, what do you guys think about this version?

apahim:
Closing as it's easier to check licensing while reviewing the code than using travis.

PR #57, opened 2016-06-09
Added fsfuzzer to avocado

PraveenPenguin:
added fsfuzzer Test case

### DIFF
--- a/fuzz/fsfuzzer.py
+++ b/fuzz/fsfuzzer.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+# Based on code by Martin Bligh <mbligh@google.com>
+#   copyright: 2006 Google
+#   https://github.com/autotest/autotest-client-tests/tree/master/fsfuzzer
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import process
+from avocado.utils import build
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Fsfuzzer(Test):
+
+    '''
+    fsfuzzer is a file system fuzzer tool. This test simply runs fsfuzzer
+    Fuzzing is slang for fault injection via random inputs. The goal is to
+    find bugs in software without reading code or designing detailed test
+    cases.
+       fsfuzz will inject random errors into the files systems
+       mounted. Evidently it has found many errors in many systems.
+    '''
+
+    def setUp(self):
+        '''
+        Build fsfuzzer
+        Source:
+        http://people.redhat.com/sgrubb/files/fsfuzzer-0.6.tar.gz
+        '''
+        sm = SoftwareManager()
+        if not sm.check_installed("gcc") and not sm.install("gcc"):
+            self.error("Gcc is needed for the test to be run")
+        tarball = self.fetch_asset('http://people.redhat.com'
+                                   '/sgrubb/files/fsfuzzer-0.6.tar.gz')
+        data_dir = os.path.abspath(self.datadir)
+        archive.extract(tarball, self.srcdir)
+        version = os.path.basename(tarball.split('.tar.')[0])
+        self.srcdir = os.path.join(self.srcdir, version)
+
+        # Patch for fsfuzzer
+        patch = self.params.get(
+            'patch', default='makefile.patch')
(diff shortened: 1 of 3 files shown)


sathnaga:
@PraveenPenguin Please provide proper description for the pull request

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,74 @@
...
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import process
+from avocado.utils import build
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Fsfuzzer(Test):
+
+    '''
```
ldoktor, line 33: nitpick: The use of `"""` is more common for normal docstrings (we used to have a picky guy who regularly replaced `'''` with `"""`). Anyway not a big deal.

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,74 @@
...
+        '''
+        Build fsfuzzer
+        Source:
+        http://people.redhat.com/sgrubb/files/fsfuzzer-0.6.tar.gz
+        '''
+        sm = SoftwareManager()
+        if not sm.check_installed("gcc") and not sm.install("gcc"):
+            self.error("Gcc is needed for the test to be run")
+        tarball = self.fetch_asset('http://people.redhat.com'
+                                   '/sgrubb/files/fsfuzzer-0.6.tar.gz')
```
ldoktor, line 52: there is a new version 0.7 which fixes lots of bugs, introduces new one and adds support for new filesystems. I think it'd be wise to use that one...

PS: You need to add patch:

``` patch
--- fstest.c.orig       2016-06-16 11:47:18.588732862 +0200
+++ fstest.c    2016-06-16 11:41:24.399070406 +0200
@@ -35,6 +35,7 @@
 #include <signal.h>
 #include <sys/file.h>
 #include <sys/mman.h>
+#include <sys/stat.h>

 #define LOGGIT 1
 #define CHECK_XATTR 1
```

in order to compile it.

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,74 @@
...
+        p1 = 'patch -p1 < %s/%s' % (data_dir, patch)
+
+        process.run(p1, shell=True)
+
+        build.make(self.srcdir)
+
+    def test(self):
+
+        args = self.params.get('fstype', default='iso9660') + ' 1'
+        process.system(self.srcdir + '/run_test ' + args)
```
ldoktor, line 71: well yes and no. This way you only run a single test of a single fs. This test should be executed by `./fsfuzz` optionally followed by a filesystem type. If none is chosen, all of them are tried.

So I'd suggest using `./fsfuzz` + `args` with default `""`.

Also this test requires root, so please use `sudo=True` argument to the `process.system`.

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,74 @@
...
+
+class Fsfuzzer(Test):
+
+    '''
+    fsfuzzer is a file system fuzzer tool. This test simply runs fsfuzzer
+    Fuzzing is slang for fault injection via random inputs. The goal is to
+    find bugs in software without reading code or designing detailed test
+    cases.
+       fsfuzz will inject random errors into the files systems
+       mounted. Evidently it has found many errors in many systems.
```
ldoktor, line 39: I think it's worth adding the warning from the fsfuzzer README:

```
WARNING: This program can cause your kernel to oops. ITS WHOLE
PURPOSE IS TO PROVOKE THE COMPUTER TO DO BAD THINGS. If you are
using a journalling file system for all your important data, for
the most part you will be OK running this on your machine. But its
impossible to know what hole in the OS you will trigger and what its
consequences are. If you don't run with a journalled file system 
and many processes are active and some writing to disk, there
is a real good chance that you could screw up your computer. YOU
HAVE BEEN WARNED!
```

And I'd also append a message:

```
Before each test a message to system-log is written to identify what test might have caused the panic.
```

fuzz/fsfuzzer.py.data/fsfuzzer.yaml
```diff
@@ -0,0 +1,4 @@
+setup:
+	filesystem : !mux
+		default : 
+			fstype : iso9660
```
ldoktor, line 4: please use soft-spaces instead of tabs. Also for this test it'd make sense to include all fstypes. For 0.7 it means:

```
$ ./fsfuzz --help
./fsfuzz cramfs|ext2|ext3|ext4|swap|iso9660|msdos|vfat|xfs
```

fuzz/fsfuzzer.py.data/makefile.patch
```diff
@@ -0,0 +1,11 @@
+--- fsfuzzer-0.6/Makefile.orig	2011-02-07 19:48:48.000000000 -0800
++++ fsfuzzer-0.6/Makefile	2011-02-07 19:48:57.000000000 -0800
+@@ -15,7 +15,7 @@
+ CFLAGS=-g -W -Wall -Wundef
+ LIBS=
+ all:
+-	gcc $(CFLAGS) mangle.c -o mangle $(LIBS)
++	$(CC) $(CFLAGS) mangle.c -o mangle $(LIBS)
```
ldoktor, line 8: this won't be needed in case we use the 0.7 version as it supports `configure`

PraveenPenguin:
@ldoktor  thanks to looking into this , will add all suggested  comments  and push new commit .

PraveenPenguin:
@ldoktor Made changes and pushed #66 
Thanks

PR #58, opened 2016-06-10
Ported disktest from autotest [v2]

Nosheen08:
Signed-off-by: Nosheen nopathan@linux.vnet.ibm.com

PR #60, opened 2016-06-14
Add io.disk.disktest test [v3]

ldoktor:
This pull request backports a `disktest` test from `Autotest`.

v1: https://github.com/avocado-framework/avocado-misc-tests/pull/28
v2: https://github.com/avocado-framework/avocado-misc-tests/pull/58

changelog:

``` yaml
v3: Added new commit with fixes I find necessary for inclusion
v3: Add new commit to remove unnecessary Makefile
v3: Added new commit to improve log analysis
```

**NOTE: The first two commits needs to be squashed together before merge**


Nosheen08:
Hi @ldoktor , thanks for the contribution. Its working fine for me. I wanted to ask you can we use self.skip() instead of self.error() here, so the status of test will be SKIP not an ERROR?

screenshot from 2016-06-14 16-21-16

Nosheen08:
Also, we need to change the parameter name "disk1" as "disks" in disktest.yaml file. We can remove "makefile" parameter as it is not needed now right ?
screenshot from 2016-06-14 16-33-34

ldoktor:
Hello @Nosheen08 please click on the patch and add comments in the place using the `+` on the line number. Using pictures makes things hard to follow.

Aynway the answer would be yes, I can use skip, I kept the `error` only because it was used in your version. I'd actually prefer `skip`, so let's fix it in v4.

Regarding the params yes, I forgot to update the sample yaml file.

Thank you for the comments, please try it out and when you feel like it's OK I'll send v4 with the first two commits squashed. (and authored to both of us)

Nosheen08:
Hello @ldoktor  , ok next time I ll take care of it. And the test is working fine for me I have checked it. So you can send v4 with the first two commits squashed. Thanks.

PR #65, opened 2016-06-17
Adding test for Module Unload/Load [V2]

hathyaga:
This Program loads and unloads the kernel driver modules along with the dependant modules for a total of 100 counts as a stress test.

v1: https://github.com/avocado-framework/avocado-misc-tests/pull/27

### DIFF
--- a/io/driver/module_unload_load.sh
+++ b/io/driver/module_unload_load.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>
+
+CONFIG_FILE="$AVOCADO_TEST_DATADIR"/config
+DRIVER=`lspci -k | grep -iw "Kernel driver in use" | cut -d ':' -f2 | sort | uniq`
+
+module_load() {
+    echo "Reloading driver $1"
+    modprobe $1
+    if [ $? != 0 ]; then
+        echo "Failed to load driver module $1"
+        break;
+    fi
+    echo
+}
+
+
+module_unload() {
+    for i in $(cat $CONFIG_FILE | grep "$1=" | awk -F'=' '{print $2}'); do
+        module_unload $i
+        if [ $? != 0 ]; then
+            return
+        fi
+    done
+    echo "Unloaded driver $1"
+    rmmod $1
+    if [ $? != 0 ]; then
+        echo "Failed to unload driver module $i"
+        break;
+    fi
+}
+
+
+for driver in $DRIVER; do
+    echo "Starting driver module load/unload test for $driver"
+    echo
+    for j in $(seq 1 100) ; do
+        module_unload $driver
+        # Sleep for 5s to allow the module unload to complete
+        sleep 5
+        module_load $driver
+        # Sleep for 5s to allow the module load to complete
+        sleep 5
+    done
+    echo "Finished driver module load/unload test for $driver"
+    echo
+done
--- a/io/driver/module_unload_load.sh.data/README.txt
+++ b/io/driver/module_unload_load.sh.data/README.txt
@@ -1,0 +1,3 @@
+This Program loads and unloads the kernel driver modules. For the kernel driver modules that have dependencies, the dependant modules should  be listed in the config file (config). The script takescare of unloading the dependant modules, if any, before unloading the core module. It runs for a total of 100 times as stress.
+
+This script should be run as root.
--- a/io/driver/module_unload_load.sh.data/config
+++ b/io/driver/module_unload_load.sh.data/config
@@ -1,0 +1,2 @@
+mlx4_core=mlx4_en mlx4_ib
+mlx5_core=mlx5_ib


ldoktor:
Hello @hathyaga, I'm sorry it took so long, you did not include the link to the original PR therefor I was not notified about this new PR.

Overall yes, you skip loading/unloading modules, which failed, but you don't reflect that in the results. It might actually happened that none of the modules will be removed and the test passes. I'd recommend failing the test when there are modules which failed to be removed, but that brings one problem, the disk driver should not be successfully removed, therefor all tests would fail. The solution would be to use blacklist and keep adding devices which are unremovable.

Other way to solve this would be to check if at least one device was added/removed properly and pass the test only in such case.

PS: On my Fedora with Fedora guest the test still fails when using virtio (which is not your problem) so I used ide and the following modules are unloadable: `ata_piix,ehci-pci,piix4_smbus,qxl,uhci_hcd`. There are few other modules which are successfully removed (`snd_hda_intel,virtio-pci`), so the second proposal would pass.

io/driver/module_unload_load.sh
```diff
@@ -0,0 +1,60 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>
+
+CONFIG_FILE="$AVOCADO_TEST_DATADIR"/config
+DRIVER=`lspci -k | grep -iw "Kernel driver in use" | cut -d ':' -f2 | sort | uniq`
```
ldoktor, line 18: `ERR=""`

io/driver/module_unload_load.sh
```diff
@@ -0,0 +1,60 @@
...
+    for i in $(cat $CONFIG_FILE | grep "$1=" | awk -F'=' '{print $2}'); do
+        module_unload $i
+        if [ $? != 0 ]; then
+            return
+        fi
+    done
+    echo "Unloaded driver $1"
+    rmmod $1
+    if [ $? != 0 ]; then
+        echo "Failed to unload driver module $i"
```
ldoktor, line 41: `ERR="$ERR,unload-$1"`

io/driver/module_unload_load.sh
```diff
@@ -0,0 +1,60 @@
...
+# Author: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>
+
+CONFIG_FILE="$AVOCADO_TEST_DATADIR"/config
+DRIVER=`lspci -k | grep -iw "Kernel driver in use" | cut -d ':' -f2 | sort | uniq`
+
+module_load() {
+    echo "Reloading driver $1"
+    modprobe $1
+    if [ $? != 0 ]; then
+        echo "Failed to load driver module $1"
```
ldoktor, line 24: `ERR="$ERR,load-$1"`

io/driver/module_unload_load.sh
```diff
@@ -0,0 +1,60 @@
...
+        module_unload $driver
+        # Sleep for 5s to allow the module unload to complete
+        sleep 5
+        module_load $driver
+        # Sleep for 5s to allow the module load to complete
+        sleep 5
+    done
+    echo "Finished driver module load/unload test for $driver"
+    echo
+done
```
ldoktor, line 60:
``` bash
if [ $ERR ]; then
    echo Some modules failed to load/unload ${ERR:1}
    exit 1
fi
```

ldoktor, line 60: alternatively you might not `exit 1` when there are any modules which passed.

io/driver/module_unload_load.sh
ldoktor, line 18: `PASS=""`

io/driver/module_unload_load.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Starting driver module load/unload test for $driver"
+    echo
+    for j in $(seq 1 100) ; do
+        module_unload $driver
+        # Sleep for 5s to allow the module unload to complete
+        sleep 5
+        module_load $driver
+        # Sleep for 5s to allow the module load to complete
+        sleep 5
+    done
```
ldoktor, line 57:
``` bash
if [ "$j" -eq 3 ]; then
    echo "Finished driver module load/unload test for $driver"
    PASS="$PASS,$driver"
else
    echo "Failed to load/unload $driver"
fi
echo
```

io/driver/module_unload_load.sh
ldoktor, line 60:
``` bash
if [ $ERR ]; then
    echo Some modules failed to load/unload ${ERR:1}
    exit 1
fi
if [ "$PASS" ]; then
    echo Successfully loaded/unloaded: ${PASS:1}
else
    exit -1
```

ldoktor:
OK I went ahead and added comments to support both of the variants, please take a look and use either of them. As for the blacklist we can add it in the future if needed.

io/driver/module_unload_load.sh
ldoktor, line 18: Actually one more thing, the `DRIVER` should be probably renamed to `DRIVERS` as it stores all drivers we're going to use, not just single driver. It's slightly confusing later in the code.

hathyaga:
Hello @ldoktor, 
Sorry for the late reply. I'm still working on this PR.
Need one clarification here:
The 2 variants that your are talking about is the ERR and PASS ?? 
and I have to be using any 1 of them?

I have checked for the ERR and it works fine. The initial code was not handling the case of what to do when the module unload/load fails. This has now been taken care of using ERR.

As for PASS and using [ "$j" -eq 3 ], I reckon this is not needed as we run the script on purpose for 100 counts. Can this be disregarded?

ldoktor:
Hello @hathyaga, the two variants belongs to: https://github.com/avocado-framework/avocado-misc-tests/pull/65#issuecomment-229263313 which means I'm fine with finishing with PASS only and only when all modules are successfully removed/added, or to PASS when at least one cycle is successful (at least one module is successfully removed/added).

As for the `-eq 3` it was in my modified version, the number should be 100 (although ideally you should define variable at the beginning and use it everywhere, something like NO_ITERATIONS)

hathyaga:
@idoktor Closing this PR. Created a new PR-->  https://github.com/avocado-framework/avocado-misc-tests/pull/151

PR #66, opened 2016-06-17
Added fsfuzzer to avocado[v2]

PraveenPenguin:
added fsfuzzer Test case

### DIFF
--- a/fuzz/fsfuzzer.py
+++ b/fuzz/fsfuzzer.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+# Based on code by Martin Bligh <mbligh@google.com>
+#   copyright: 2006 Google
+#   https://github.com/autotest/autotest-client-tests/tree/master/fsfuzzer
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import process
+from avocado.utils import build
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Fsfuzzer(Test):
+
+    """
+    fsfuzzer is a file system fuzzer tool. This test simply runs fsfuzzer
+    Fuzzing is slang for fault injection via random inputs. The goal is to
+    find bugs in software without reading code or designing detailed test
+    cases.
+       fsfuzz will inject random errors into the files systems
+       mounted. Evidently it has found many errors in many systems.
+    """
+
+    def setUp(self):
+        '''
+        Build fsfuzzer
+        Source:
+        http://people.redhat.com/sgrubb/files/fsfuzzer-0.7.tar.gz
+        '''
+        sm = SoftwareManager()
+        if not sm.check_installed("gcc") and not sm.install("gcc"):
+            self.error("Gcc is needed for the test to be run")
+        tarball = self.fetch_asset('http://people.redhat.com'
+                                   '/sgrubb/files/fsfuzzer-0.7.tar.gz')
+        data_dir = os.path.abspath(self.datadir)
+        archive.extract(tarball, self.srcdir)
+        version = os.path.basename(tarball.split('.tar.')[0])
+        self.srcdir = os.path.join(self.srcdir, version)
+
+        # Patch for fsfuzzer
+        patch = self.params.get(
+            'patch', default='fstestfix.patch')
+        patch_1 = self.params.get(
+            'patch', default='ppc64leconfig_guess.patch')
+        patch_2 = self.params.get(
+            'patch', default='attrfix.patch')
+        patch_3 = self.params.get(
+            'patch', default='fsfuzz_fix.patch')
+        os.chdir(self.srcdir)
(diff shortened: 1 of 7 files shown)


fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,88 @@
...
+    cases.
+       fsfuzz will inject random errors into the files systems
+       mounted. Evidently it has found many errors in many systems.
+    """
+
+    def setUp(self):
+        '''
+        Build fsfuzzer
+        Source:
+        http://people.redhat.com/sgrubb/files/fsfuzzer-0.7.tar.gz
```
ldoktor, line 46: Hey, I talked to Steve and the official `fsfuzzer` is now located at https://github.com/stevegrubb/fsfuzzer and he applied the import patch. So perhaps you could send all the patches as PR to that repo so eventually we don't need to use any workarounds at all, only the latest master:

https://github.com/stevegrubb/fsfuzzer/archive/master.zip

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,88 @@
...
+        tarball = self.fetch_asset('http://people.redhat.com'
+                                   '/sgrubb/files/fsfuzzer-0.7.tar.gz')
+        data_dir = os.path.abspath(self.datadir)
+        archive.extract(tarball, self.srcdir)
+        version = os.path.basename(tarball.split('.tar.')[0])
+        self.srcdir = os.path.join(self.srcdir, version)
+
+        # Patch for fsfuzzer
+        patch = self.params.get(
+            'patch', default='fstestfix.patch')
```
ldoktor, line 60: Do you expect the `patch` to be changed regularly from yaml? If not then it should be only hardcoded...

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,88 @@
...
+        patch_1 = self.params.get(
+            'patch', default='ppc64leconfig_guess.patch')
```
ldoktor, line 62: Actually all of these only use `patch` key from params, so when `patch` is set in `yaml` it applies the same patch 4 times, which does not make sense (and it fails)

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,88 @@
...
+        p1 = 'patch -p1 < %s/%s' % (data_dir, patch)
+        p2 = 'patch -p1 < %s/%s' % (data_dir, patch_1)
+        p3 = 'patch -p1 < %s/%s' % (data_dir, patch_2)
+        p4 = 'patch -p1 < %s/%s' % (data_dir, patch_3)
+
+        process.run(p1, shell=True)
+        process.run(p2, shell=True)
+        process.run(p3, shell=True)
+        process.run(p4, shell=True)
+        process.run('[ -x configure ] && ./configure', shell=True)
```
ldoktor, line 77: Well all new versions should have the `configure` so I don't think it's useful to check for it....

PraveenPenguin:
@ldoktor I don't think these patch's going to change regularly, if you say i can hardcode those patches ?

fuzz/fsfuzzer.py
```diff
@@ -0,0 +1,88 @@
...
+
+        build.make(self.srcdir)
+
+    def test(self):
```
ldoktor, line 81: It'd be nice to add docstring explaining the scenario. Something like:

```
Runs the fsfuzz test suite. By default uses all supported fstypes,
but you can specify only one by `fstype` param.
```

fuzz/fsfuzzer.py.data/attrfix.patch
```diff
@@ -0,0 +1,20 @@
+diff -crB a/fstest.c b/fstest.c
```
ldoktor, line 1: Where these patches comes from? Would you please send them to https://github.com/stevegrubb/fsfuzzer so eventually we can remove them?

fuzz/fsfuzzer.py.data/fsfuzzer.yaml
```diff
@@ -0,0 +1,4 @@
+setup:
+ filesystem : !mux
+  default : 
+   fstype : cramfs|ext2|ext3|ext4|swap|iso9660|msdos|vfat|xfs
```
ldoktor, line 4: This is not supported fstype. I only pointed out what are the supported ones. What I had in mind is:

```
!mux
cramfs:
    fstype: cramfs
ext2:
    fstype: ext2
ext3:
    fstype: ext3
ext4:
    fstype: ext4
swap:
    fstype: swap
iso9660:
    fstype: iso9660
msdos:
    fstype: msdos
vfat:
    fstype: vfat
xfs:
    fstype: xfs
```

Which when executed would run all types as separate avocado tests.

ldoktor, line 4: Actually there are more:

```
mksquashfs
hfs
gfs2
jffs2
reiserfs
romfs
ecryptfs
```

These were not supported on my machine, but they are supported by the fsfuzz.

So I'd recommend adding all of them and use `./fsfuzz  --help` to check if the current one is supported during `setUp` and use `self.skip` in case it's not supported.

PraveenPenguin:
@ldoktor  thanks , will implement all comments and submit  code for review .

can we wait till when fsfuzzer patches got merged in githug fsfuzzer git repo  or will send a modified code ?

ldoktor:
I'd prefer to send the patches to fsfuzzer and new version of this test without the patches. Steve merged my patch really quickly so hopefully he'll do the same with yours.

PraveenPenguin:
@ldoktor Steve still need to merge the code i sent him pull request 10 days back , can you please check with him about the this 
or can i commit the code with patch files 

Thanks

ldoktor:
Hello @PraveenPenguin I talked to Steve and he promised to take a look. He wants to use `autogen` instead of custom config script.

ldoktor:
Hello @PraveenPenguin the fsfuzz was updated. The workflow is `./autogen.sh && ./configure && make`. Would you please address the issues from this PR and this little modification and push new version?

PraveenPenguin:
@ldoktor sure i will create new PR with updated code .

Thanks

PraveenPenguin:
@ldoktor made change and pushed #92  , Please have look
Thanks

PR #67, opened 2016-06-17
Adding lvsetup Test for Avocado [v2]

harish-24:
Changes from #44 
1. Change in work-flow of test
2. Added setUp and tearDown for test

### DIFF
--- a/io/disk/lvsetup.py
+++ b/io/disk/lvsetup.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harish <harisrir@linux.vnet.ibm.com>
+#
+# Based on the code by:
+#
+# Copyright: 2012 Intra2net
+# Author: Plamen Dimitrov (plamen.dimitrov@intra2net.com)
+
+"""
+Test that automatically takes shapshots from created logical volumes
+using a given policy.
+
+For details about the policy see README.
+"""
+import avocado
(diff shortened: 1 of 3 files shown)


io/disk/lvsetup.py
```diff
@@ -0,0 +1,120 @@
...
+        lv_utils.lv_take_snapshot(self.vg_name, self.lv_name,
+                                  self.lv_snapshot_name,
+                                  self.lv_snapshot_size)
+        lv_utils.lv_revert(self.vg_name, self.lv_name, self.lv_snapshot_name)
+
+    def tearDown(self):
+        """
+        Clear all PV,VG, LV and snapshots created by the test.
+        """
+        # Remove created VG
```
ldoktor: Hello @harish-24 this is much better, thanks. Still the cleanup is not complete and it could be actually way simpler, you could simply use `vg_ramdisk_cleanup` function, which tries to gracefully cleanup the system and raises the exception after this is done.

If, for some reason, you don't want to use this function, please add `umount` for loop device and `self.error` at the end of the cleanup if any of cleanups failed as we don't want to affect following tests by failed cleanup without a notice (and message in log is not sufficient, the test needs to fail and avocado actually notifies you that it was the cleanup what failed).

io/disk/lvsetup.py.data/lvsetup.yaml
```diff
@@ -0,0 +1,6 @@
+scenario: !mux
+    loopback:
+        disk: None
+        lv_size: "10M"
+    real:
+        disk: /dev/sde
```
ldoktor: I'd strongly suggest commenting this second one out to avoid people complaining about wiped drives...

```
# Uncomment this to use real drive (warning: it'll be erased!)
# real:
#    disk: /dev/sde

And just to be sure, you might actually use `/dev/vde` to avoid unnecessary corruption of main drive...
```

io/disk/lvsetup.py
harish-24: The `vg_ramdisk_cleanup` process seems to be not cleaning created loop device and PV as well. So thought of handling each individually. Will add `umount` and `self.error` with necessary comments. Thanks.

io/disk/lvsetup.py.data/lvsetup.yaml
harish-24: Sure

harish-24:
@ldoktor Have updated with the changes. Thanks.

PR #71, opened 2016-06-20
Added pi_tests test  to avocado

PraveenPenguin:
added pi_tests Test case


ldoktor:
Hello @PraveenPenguin this test is part of the https://github.com/avocado-framework/avocado-misc-tests/pull/54 which is under review. I think it'd be better to include it into the yaml file there instead of building yet another test, what do you think?

PraveenPenguin:
@ldoktor yes that make sense .

ldoktor:
OK, feel free to join the discussion there on the default arguments...

PR #59, opened 2016-06-14
Added testcase for network tools like Arp,Netstat etc.,

basheerk:
Added testcase to verify the functionality of various network tools like
ifconfig,hostname,arp,traceroute,route,ipmaddr,iptunnel,netstat.

Signed-off-by : Basheer K basheer@linux.vnet.ibm.com
Tested-by     : Basheer K basheer@linux.vnet.ibm.com

PR #74, opened 2016-06-22
[PATCH V2]Added testcase to verify the functionality of various network tools.

basheerk:
Test to verify the functionality network utilities like ifconfig,hostname,arp,traceroute,route,ipmaddr,iptunnel,netstat.

Signed-off-by : Basheer K basheer@linux.vnet.ibm.com
Tested-by     : Basheer K basheer@linux.vnet.ibm.com

v1: https://github.com/avocado-framework/avocado-misc-tests/pull/59

### DIFF
--- a/io/net/net_tools.py
+++ b/io/net/net_tools.py
@@ -1,0 +1,359 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Basheer K<basheer@linux.vnet.ibm.com>
+#
+# Based on code by Hong Bo Peng <penghb@cn.ibm.com>
+# copyright: 2003, 2015 IBM Corp
+
+import os
+import re
+import avocado
+from avocado import Test
+from avocado import main
+from avocado.utils import process, distro
+from avocado.utils.software_manager import SoftwareManager
+
+
+def install_dependencies():
+    """
+    To check and install dependencies for the test
+    """
+    detected_distro = distro.detect()
+    sm = SoftwareManager()
+    if detected_distro.name == "SuSE":
+        net_tools = ("net-tools", "traceroute")
+    else:
+        net_tools = ("net-tools", "hostname", "traceroute")
+    for pkg in net_tools:
+        if not sm.check_installed(pkg) and not sm.install(pkg):
+            raise AssertionError("%s package is need to test" % pkg)
+
+
+class Hostname(Test):
+    """
+    hostname: set/get hostname by cmd. Check it with the expected result.
+    Test to verify the functionality of the hostname command
+    """
+
+    def setUp(self):
+        install_dependencies()
+        self.restore_hostname = False
+        # Get Hostname
+        hostname = process.system_output("hostname").strip('\n')
+        if not hostname:
+            # set hostname if not set
+            process.system("hostname localhost.localdomain", sudo=True)
+            hostname = "localhost.localdomain"
+        self.hostname = hostname
+
+    @avocado.fail_on(process.CmdError)
+    def test_hostname(self):
+        """
+        i.Verifies hostname command options and
+        ii.Test to change the hostname
+        """
+        output = process.system_output("hostname")
+        if not output:
+            self.fail("unexpected response from hostname command")
+
+        # Verifying different options provided by hostname
+        options_to_verify = self.params.get('hostname_opt', default="f")
+        for option in options_to_verify:
+            ret = process.run("hostname -%s" % option, ignore_status=True)
+            if ret.exit_status:
+                self.fail("Hostname reported non-zero status %s for option %s"
+                          % (ret.exit_status, option))
+            if not ret.stdout:
+                self.fail("No output for %s option" % (option))
+
+        # Test to change hostname
+        myhostname_file = os.path.join(self.workdir, "MYHOSTNAME")
+        myhostname = "myhost.my-domain"
+        if myhostname == self.hostname:
+            myhostname += '1'
+        with open(myhostname_file, 'w') as fobj:
+            fobj.write(myhostname)
+
+        process.system("hostname -F %s" % myhostname_file,
+                       sudo=True)
+        self.restore_hostname = True
+        if myhostname not in process.system_output(
+                "hostname", env={"LANG": "C"}):
+            self.fail("unexpected response from hostname -F command and " +
+                      "hostname -F didn't set hostname")
+
+    def tearDown(self):
+        if self.restore_hostname:
+            # Restore hostname
+            process.system("hostname %s" % self.hostname,
+                           sudo=True)
+            if self.hostname not in process.system_output(
+                    "hostname", env={"LANG": "C"}):
+                self.error("Failed to restore Hostname")
+
+
+class Ifconfig(Test):
+    """
+    ifconfig - configure a network interface
+    Test to verify the ifconfig functionality.
+    """
+
+    def setUp(self):
+        if process.system("ping -c 5 -w 5 localhost", ignore_status=True):
+            self.skip("Unable to ping localhost")
+        self.lo_up = True
+        if "lo:1" in process.system_output("ifconfig", env={"LANG": "C"}):
+            self.skip("alias for an loopback interface is configured")
+        self.alias = None
+        install_dependencies()
+        self.ipv6 = False
+        if os.path.exists("/proc/net/if_inet6"):
+            self.ipv6 = True
+
+    @avocado.fail_on(process.CmdError)
+    def test_ifconfig(self):
+        """
+        Verify the functionality of the ifconfig
+        """
+        output = process.run("ifconfig", env={"LANG": "C"})
+        if "Local Loopback" not in output.stdout:
+            self.fail("unexpected output of ifconfig")
+        if self.ipv6:
+            if "inet6" not in output.stdout:
+                self.fail("Did not see IPV6 info")
+        # setup and verify an alias interface
+        self.alias = "lo:1"
+        process.system("ifconfig %s 127.0.0.240 netmask 255.0.0.0" %
+                       self.alias,
+                       sudo=True)
+        if "lo:1" not in process.system_output("ifconfig", env={"LANG": "C"}):
+            self.fail("Failed to configure alias for loopback")
+
+        self._remove_alias(self.alias)
+        self.alias = None
+        # Test to make loopback interface up and down
+        process.system("ifconfig lo down", sudo=True)
+        self.lo_up = False
+        if not process.system("ping -c 5 -w 5 localhost", ignore_status=True):
+            self.fail("Failed to change loopback interface to down state")
+
+    @staticmethod
+    def _remove_alias(name):
+        process.system("ifconfig %s down" % name,
+                       sudo=True)
+        if "lo:1" in process.system_output("ifconfig", env={"LANG": "C"}):
+            raise AssertionError("Failed to remove alias for loopback")
+
+    @staticmethod
+    def _restore_lo_intf_state():
+        process.system("ifconfig lo up", sudo=True)
+        if process.system("ping -c 5 -w 5 localhost", ignore_status=True):
+            raise AssertionError("Failed to restore loopback interface state")
+
+    def tearDown(self):
+        errs = []
+        if self.alias:
+            try:
+                self._remove_alias(self.alias)
+            except Exception:
+                errs.append("remove alias")
+        if not self.lo_up:
+            try:
+                self._restore_lo_intf_state()
+            except Exception:
+                errs.append("restore lo state")
+        if errs:
+            self.error("Failed to: %s" % ", ".join(errs))
+
+
+class Arp(Test):
+    """
+    arp: run it with different options. Check the return code.
+    """
+
+    def setUp(self):
+        interface_out = process.system_output("ip route show default",
+                                              env={"LANG": "C"})
+        if "default via" not in interface_out:
+            self.skip("No active interface with deafult gateway configured")
+        install_dependencies()
+        search_obj = re.search(
+            "^default via\s+(\S+)\s+dev\s+(\w+)", interface_out)
+        self.default_router = search_obj.group(1)
+
+    @avocado.fail_on(process.CmdError)
+    def test_arp(self):
+        """
+        Test to resolve Mac addr of default gateway router using arp
+        """
+        process.system("ping -c 2 -w 5 %s" % self.default_router)
+        output = process.run(cmd="arp -n", ignore_status=True,
+                             env={"LANG": "C"})
+        if output.exit_status:
+            self.fail("Arp reported non zero exit status")
+        if self.default_router not in output.stdout:
+            self.fail("unexpected response from arp")
+
+    def tearDown(self):
+        pass
+
+
+class NetworkUtilities(Test):
+    """
+    traceroute,traceroute6: run for localhost. It should report 1 hop.
+    route: run it with -n. Check the return code.
+    ipmaddr: run it and check the return code.
+    """
+
+    def setUp(self):
+        install_dependencies()
+        self.ipv6 = False
+        if os.path.exists("/proc/net/if_inet6"):
+            self.ipv6 = True
+
+    @avocado.fail_on(process.CmdError)
+    def test_traceroute(self):
+        """
+        Verify traceroute,traceroute6 functionality.
+        """
+        ret = process.run(cmd="traceroute localhost",
+                          env={"LANG": "C"})
+        no_of_hops = re.search(
+            "(\d+)\s+\S+\s*\(127.0.0.1\)",
+            ret.stdout).group(1)
+        # Only one hop is required to get to localhost.
+        if str(no_of_hops) != '1':
+            self.fail("traceroute did not show 1 hop for localhost")
+
+        if self.ipv6:
+            detected_distro = distro.detect()
+            if detected_distro.name in ("SuSE", "Ubuntu"):
+                ret = process.run(cmd="traceroute6 ipv6-localhost",
+                                  env={"LANG": "C"})
+            else:
+                ret = process.run(cmd="traceroute6 localhost6",
+                                  env={"LANG": "C"})
+            no_of_hops = re.search(
+                "(\d+)\s+\S+\s*\(::1\)",
+                ret.stdout).group(1)
+            if str(no_of_hops) != '1':
+                self.fail(
+                    "traceroute6 did not show 1 hop for " +
+                    "localhost6/ipv6-localhost")
+
+    @avocado.fail_on(process.CmdError)
+    def test_netstat(self):
+        """
+        Verify the functionality of netstat
+        """
+        # Verifying different options provided by hostname
+        options_to_verify = self.params.get('netstat_opt', default="s")
+        for option in options_to_verify:
+            ret = process.run("netstat -%s" % option, verbose=False,
+                              ignore_status=True)
+            if ret.exit_status:
+                self.fail(
+                    "Netstat command reported non-zero status %s " %
+                    ret.exit_status +
+                    "for option %s" %
+                    option)
+
+    @avocado.fail_on(process.CmdError)
+    def test_route(self):
+        """
+        To verify route command utility
+        """
+        ret = process.run(cmd="route -n", ignore_status=True)
+        if ret.exit_status:
+            self.fail(
+                "route command reported non-zero %s exit status" %
+                ret.exit_status)
+        if self.ipv6:
+            ret = process.run(cmd="route -A inet6 -n", ignore_status=True)
+            if ret.exit_status:
+                self.fail(
+                    "route command reported non-zero %s exit status " %
+                    ret.exit_status +
+                    "while displaying ipv6 route table")
+
+    @avocado.fail_on(process.CmdError)
+    def test_ipmaddr(self):
+        """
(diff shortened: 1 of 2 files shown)


basheerk:
@ldoktor updated V2 for network tools test

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def setUp(self):
+        install_dependencies()
+        self.restore_hostname = False
+        # Get Hostname
+        output = process.run("hostname", verbose=False,
+                             env={"LANG": "C"})
+        if not output.stdout:
+            # set hostname if not set
+            if not process.system("hostname localhost.localdomain", sudo=True):
```
ldoktor: `process.system` raises exception on failure. You need to use `ignore_status=True` to be able to check for failure. (or simply use `process.system()` without the `checking` as the test finishes with `ERROR` when this command fails...)

ldoktor: Btw the later would be IMO better as you get details about the failure.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+                self.error("Failed to set hostname\n")
+
+        self.hostname = process.run("hostname", verbose=False,
+                                    env={"LANG": "C"}).stdout
```
ldoktor: This second call is only needed when the hostname was updated. So how about:

``` python
self.restore_hostname = False
hostname = process.system_output("hostname", verbose=False)
if not hostname:
    process.system("hostname localhost.localdomain", sudo=True)
    hostname = "localhost.localdomain"
self.hostname = hostname
```

PS: The `env={"LANG": "C"}` is only important for outputs where you match the output, it does not have to be everywhere.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        # Backup info required to restore hostname
+        self.workdir = data_dir.get_tmp_dir()
```
ldoktor: `self.workdir` already exists in `Test`, you don't have to initialize it.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        self.backup_dir = os.path.join(self.workdir, 'net-tools')
```
ldoktor: The `self.workdir` is already your tmpdir, I don't see a point in creating yet another directory in it.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        fobj.write("myhost.my-domain")
+        fobj.close()
+
+    def test_hostname(self):
+        """
+        i.Verifies hostname command options and
+        ii.Test to change the hostname
+        """
+        output = process.run("hostname", verbose=False,
+                             env={"LANG": "C"})
```
ldoktor: no need for `LANG` here...

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        if not output.stdout:
+            self.error("unexpected response from hostname command")
+
+        # Verifying different options provided by hostname
+        options_to_verify = self.params.get('hostname_opt', default="f")
+        for option in options_to_verify:
+            output = process.run("hostname -%s" % option, verbose=False)
+            if not output.stdout:
+                self.error("unexpected response from hostname command")
```
ldoktor: Well the `process.run` raises exception on non-zero return and here you're using `error`. Both finish the test with `ERROR` status, which is reserved for generic test-unrelated errors. More suitable would be to use `self.fail` as this is actually the item under testing. (the difference is that theoretically `ERROR` should be reviewed by the test writer, while `FAIL` only means this feature does not work as expected).

Additionally this is the most important part of the test, so I'd suggest keeping `verbose=True` to be able to analyze the failures.

So how about this:

``` python
result = process.run("hostname -%s" % option, ignore_status=True)
if result.exit_code:
    self.fail("Hostname reported non-zero status %s for option %s" % (result.exit_code, option))
elif not result.stdout:
    self.fail("No output for %s option" % (option))
```

Note: I don't have to include the results in the exception as it's already in the log. Alternatively you can use the `verbose=False` but then you'd have to include the whole `result` in the `fail` message, otherwise there is no way to analyze what went wrong.

ldoktor: Well the `process.run` raises exception on non-zero return and here you're using `error`. Both finish the test with `ERROR` status, which is reserved for generic test-unrelated errors. More suitable would be to use `self.fail` as this is actually the item under testing. (the difference is that theoretically `ERROR` should be reviewed by the test writer, while `FAIL` only means this feature does not work as expected).

Additionally this is the most important part of the test, so I'd suggest keeping `verbose=True` to be able to analyze the failures.

So how about this:

```python
result = process.run("hostname -%s" % option, ignore_status=True)
if result.exit_code:
    self.fail("Hostname reported non-zero status %s for option %s" % (result.exit_code, option))
elif not result.stdout:
    self.fail("No output for %s option" % (option))
```

Note: I don't have to include the results in the exception as it's already in the log. Alternatively you can use the `verbose=False` but then you'd have to include the whole `result` in the `fail` message, otherwise there is no way to analyze what went wrong.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        self.workdir = data_dir.get_tmp_dir()
+        self.backup_dir = os.path.join(self.workdir, 'net-tools')
+        if not os.path.isdir(self.backup_dir):
+            os.makedirs(self.backup_dir)
+
+        self.hostname_file = os.path.join(self.backup_dir, "HOSTNAME")
+        self.myhostname_file = os.path.join(self.backup_dir, "MYHOSTNAME")
+        fobj = open(self.hostname_file, 'w')
+        fobj.write(self.hostname)
+        fobj.close()
```
ldoktor: Well you can restore the original hostname even without the file simply from the cmdline. Anyway if you find this safer, then I'd only improve the file handling:

``` python
with open(self.hostname_file) as fobj:
    fobj.write(self.hostname)
```

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        fobj = open(self.myhostname_file, 'w')
```
ldoktor: this `myhostname_file` is only useful for the test, so it should not be initialized here, see the details in the test method...

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+
+        # Verifying different options provided by hostname
+        options_to_verify = self.params.get('hostname_opt', default="f")
+        for option in options_to_verify:
+            output = process.run("hostname -%s" % option, verbose=False)
+            if not output.stdout:
+                self.error("unexpected response from hostname command")
+            self.log.info("verified -%s option of hostname command" % option)
+
+        # Test to change hostname
```
ldoktor, line 81: I think you should generate the `myhostname` file here and it's also necessary to use unique name. How about this:

``` python
myhostname = "myhost.my-domain"
myhostname_file = os.path.join(self.workdir, "MYHOSTNAME")
with open(myhostname_file, "w") as fobj:
    fobj.write(myhostname)
self.restore_hostname = True
process.system("hostname -F %s" % myhostname_file)
output = process.system_output("hostname")
if output != myhostname:
    self.fail(...)
```

io/net/net_tools.py
ldoktor: Actually there are lot of `process.run`s which are basically equal to test failure. So instead of treating each one I'd recommend using fail_on decorator:

``` python
import avocado
...
@avocado.fail_on(process.CmdError)
def test_hostname(self):
    ...
    if not process.system_output("hostname -%s" % option):
        self.fail("No output for %s option" % option)
```

And similarly for other `process` executions. The `fail_on` decorator turns the `process.CmdError` exceptions into test failures.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        process.system("hostname -F %s" % self.myhostname_file,
+                       verbose=False,
+                       sudo=True)
+        self.restore_hostname = True
+        output = process.run("hostname", verbose=False,
+                             env={"LANG": "C"})
+        if "myhost.my-domain" not in output.stdout:
+            self.error("unexpected response from hostname -F command and " +
+                       "hostname -F didn't set hostname")
+        ret = self._restore_hostname(self.hostname_file, self.hostname)
```
ldoktor: Do you find the `restore_hostname` part of the test? If not then it can be only in `tearDown`.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        if ret:
+            self.error(ret)
+        self.restore_hostname = False
+
+    @staticmethod
+    def _restore_hostname(hostnamefile, hostname):
```
ldoktor: Well if you want this function only to work as `restore_hostname` then I think it should not accept the arguments and it should use `self` to get the correct values. On the other hand you could also call this `set_hostname_file` and it can be used even for setting the hostname :wink:.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        # Restore hostname
+        process.system("hostname -F %s" % hostnamefile,
+                       verbose=False)
+        output = process.run("hostname", verbose=False,
+                             env={"LANG": "C"})
+        if hostname not in output.stdout:
+            return "Failed to restore Hostname"
```
ldoktor: How about raising the exception here instead of returning a message? Then you won't need any custom handling in the main body...

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def tearDown(self):
+        if self.restore_hostname:
+            ret = self._restore_hostname(self.hostname_file, self.hostname)
```
ldoktor: Yep, if you modify the `_restore_hostname` to raise exception, this basically becomes `self._restore_hostname()`...

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+            if ret:
+                self.error(ret)
+        shutil.rmtree(self.backup_dir)
```
ldoktor: No need to remove the `backup_dir`, nor `workdir`. The `workdir` is cleaned automatically by avocado.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+
+class Ifconfig(Test):
+    """
+    ifconfig - configure a network interface
+    Test to verify the ifconfig functionality.
+    """
+
+    def setUp(self):
+        if "0% packet loss" not in process.system_output(
+                "ping -c 5 -w 5 localhost", verbose=False):
```
ldoktor: I played with it and how about simpler check:

``` python
if process.system("ping -c 1 -w 5 localhost", ignore_status=True):
    self.skip(...)
```

when I disable lo it skips the test, when I enable it it proceeds.

ldoktor: Note that I'd prefer using `verbose=True`, but if you want to use `verbose=False` then I'd rather suggest:

```
result = process.run(ping -c 1 -w 5 localhost", ignore_status=True):
if result.exit_code:
    self.skip("Loopback interface not accessible:\n%s" % result)
```

That way you can check what went wrong as the ping output could be helpful.

io/net/net_tools.py
ldoktor, line 81: I forgot what I had in mind. First you should check if the hostname is already like that. So instead of `myhostname = "myhost.my-domain"` there should be:

``` python
myhostname = "myhost.my-domain"
if myhostname == self.hostname:
    myhostname += "2"
```

That way the myhostname is always different thant the original hostname.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+            self.ipv6 = True
+
+    def test_ifconfig(self):
+        """
+        Verify the functionality of the ifconfig
+        """
+        output = process.run("ifconfig", verbose=False,
+                             env={"LANG": "C"})
+        if "Local Loopback" not in output.stdout:
+            self.error("unexpected output of ifconfig")
```
ldoktor: self.fail. Please use it everywhere else when the failure is related to the tested component.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+        self.lo_intf_state = "Up"
+        if "lo:1" in process.system_output("ifconfig", verbose=False):
+            self.skip("alias for an loopback interface is configured")
+        self.alias = None
+        install_dependencies()
+        self.ipv6 = False
+        if os.path.exists("/proc/net/if_inet6"):
+            self.ipv6 = True
+
+    def test_ifconfig(self):
```
ldoktor, line 126: All of the process exceptions are actually test failures, so please use the `fail_on` decorator here as well.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+            self.error("unexpected output of ifconfig")
+        if self.ipv6:
+            if "inet6" not in output.stdout:
+                self.error("Did not see IPV6 info")
+        # setup and verify an alias interface
+        self.alias = "lo:1"
+        process.system("ifconfig lo:1 127.0.0.240  netmask 255.0.0.0",
+                       verbose=False,
+                       sudo=True)
+        output = process.run("ifconfig", verbose=False,
```
ldoktor: And please reconsider the `verbose=False`. It's useful for some setUps, but in most cases it contains vital information to analyze failures.

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+                             env={"LANG": "C"})
+        if "lo:1" not in output.stdout:
+            self.error("Failed to configure alias for loopback")
```
ldoktor: ^^  regarding the `verbose=False`, you can keep it disabled to shrink the log, but then you ought to log the whole results in case of error (or failure)

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def tearDown(self):
+        if self.alias:
+            ret = self._remove_alias(self.alias)
+            if ret:
+                self.error(ret)
+        if self.lo_intf_state is not "Up":
+            ret = self._restore_lo_intf_state()
+            if ret:
+                self.error(ret)
```
ldoktor: The `tearDown` has to always complete all the phases and raise the exception at the end. So how about:

``` python
errs = []
if self.alias:
    try:
        self._remove_alias(self.alias)
    except Exception:
        errs.append("remove alias")
if self.lo_intf_state is not "Up":
    try:
        self._restore_lo_intf_state()
    except Exception:
        errs.append("restore lo state")
if errs:
    self.error("Failed to: %s" % ", ".join(errs))
```

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+                                       env={"LANG": "C"})
+        if "0% packet loss" not in output:
+            return "loopback intf is not coming up"
+
+    def tearDown(self):
+        if self.alias:
+            ret = self._remove_alias(self.alias)
+            if ret:
+                self.error(ret)
+        if self.lo_intf_state is not "Up":
```
ldoktor: nitpick: You only check for `Up`, so how about calling it `self.lo_up` and use True/False?

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+
+    @staticmethod
+    def _remove_alias(name):
+        process.system("ifconfig %s down" % name,
+                       verbose=False,
+                       sudo=True)
+        output = process.run("ifconfig", verbose=False,
+                             env={"LANG": "C"})
+        if "lo:1" in output.stdout:
+            return "Failed to remove alias for loopback"
```
ldoktor: as before I'd suggest raising exception in case of failure (either `raise ...` or `self.fail/error` or even `assert ...`)

io/net/net_tools.py
```diff
@@ -0,0 +1,374 @@
...
+                                              verbose=False,
+                                              env={"LANG": "C"})
+        if "default via" not in interface_out:
+            self.skip("No active interface with deafult gateway configured")
+        install_dependencies()
+        search_obj = re.search(
+            "^default via\s+(\S+)\s+dev\s+(\w+)", interface_out)
+        self.default_router = search_obj.group(1)
+
+    def test_arp(self):
```
ldoktor, line 197: again, the `@fail_on` would be appreciated.

ldoktor, line 197: ... basically for all tests in here...

ldoktor:
Hello @basheerk I can see a huge improvement from the previous version. There are few patterns repeated which are not really welcomed. Please take a look and adjust all occurrences. Mainly I'm talking about the `process` (verbose, env, fail_on, ignore_status, ...), the use of `self.error` for test failures and custom handling of things that avocado already does for you.

basheerk:
@ldoktor updated the comments provided,could you please check?

narasimhan-v:
@basheerk Have you considered the deprecation of `net-tools` in favour of `iproute2` ?

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+        output = process.system_output("hostname")
+        if not output:
+            self.fail("unexpected response from hostname command")
+
+        # Verifying different options provided by hostname
+        options_to_verify = self.params.get('hostname_opt', default="f")
+        for option in options_to_verify:
+            ret = process.run("hostname -%s" % option, ignore_status=True)
+            if ret.exit_status:
+                self.fail("Hostname reported non-zero status %s for option %s"
```
ldoktor, line 76: You ought to log the full `ret` to see what was executed and what was the result including stdout/stderr, otherwise you have no hints of why it reported that exit status.

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+                          % (ret.exit_status, option))
+            if not ret.stdout:
+                self.fail("No output for %s option" % (option))
```
ldoktor: the same here

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+        myhostname_file = os.path.join(self.workdir, "MYHOSTNAME")
+        myhostname = "myhost.my-domain"
+        if myhostname == self.hostname:
+            myhostname += '1'
+        with open(myhostname_file, 'w') as fobj:
+            fobj.write(myhostname)
+
+        process.system("hostname -F %s" % myhostname_file,
+                       sudo=True)
+        self.restore_hostname = True
```
ldoktor, line 91: You should set the `restore_hostname` before executing the `hosname -F`, because if the command changes the name and crashes, you don't get to restore it... (basically please swap the lines 89+90 with 91)

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+    ifconfig - configure a network interface
+    Test to verify the ifconfig functionality.
+    """
+
+    def setUp(self):
+        if process.system("ping -c 5 -w 5 localhost", ignore_status=True):
+            self.skip("Unable to ping localhost")
+        self.lo_up = True
+        if "lo:1" in process.system_output("ifconfig", env={"LANG": "C"}):
+            self.skip("alias for an loopback interface is configured")
```
ldoktor, line 118: nitpick: ... interface is _already_ configured

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+            self.ipv6 = True
+
+    @avocado.fail_on(process.CmdError)
+    def test_ifconfig(self):
+        """
+        Verify the functionality of the ifconfig
+        """
+        output = process.run("ifconfig", env={"LANG": "C"})
+        if "Local Loopback" not in output.stdout:
+            self.fail("unexpected output of ifconfig")
```
ldoktor, line 132: please list the `output` to be able to debug the failure

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+        if self.ipv6:
+            if "inet6" not in output.stdout:
+                self.fail("Did not see IPV6 info")
```
ldoktor: the same here, please list he `output`

io/net/net_tools.py
ldoktor, line 132: actually it's `verbose` so please ignore this comment...

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+    """
+
+    def setUp(self):
+        interface_out = process.system_output("ip route show default",
+                                              env={"LANG": "C"})
+        if "default via" not in interface_out:
+            self.skip("No active interface with deafult gateway configured")
+        install_dependencies()
+        search_obj = re.search(
+            "^default via\s+(\S+)\s+dev\s+(\w+)", interface_out)
```
ldoktor, line 193: This would look nicer:

```
        search_obj = re.search("^default via\s+(\S+)\s+dev\s+(\w+)",
                               interface_out)
```

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+        """
+        Verify the functionality of netstat
+        """
+        # Verifying different options provided by hostname
+        options_to_verify = self.params.get('netstat_opt', default="s")
+        for option in options_to_verify:
+            ret = process.run("netstat -%s" % option, verbose=False,
+                              ignore_status=True)
+            if ret.exit_status:
+                self.fail(
```
ldoktor, line 267: this is quite ugly, you can break the statements:

```
    self.fail("Foo...,"
              "bar %s,"
              "baz %s"
              % (1, 2))
```

which then becomes `Foo...,bar 1,baz2` and looks much nicer

io/net/net_tools.py
```diff
@@ -0,0 +1,359 @@
...
+    @avocado.fail_on(process.CmdError)
+    def test_route(self):
+        """
+        To verify route command utility
+        """
+        ret = process.run(cmd="route -n", ignore_status=True)
+        if ret.exit_status:
+            self.fail(
+                "route command reported non-zero %s exit status" %
+                ret.exit_status)
```
ldoktor, line 282: the same here, please can you fix the formatting everywhere? (It's hard to read it this way)

io/net/net_tools.py
ldoktor, line 76: actually not, you removed the `verbose=False`. Please ignore this...

ldoktor:
OK, there are only few little issues, so let me run some functional tests and then I can merge this and send the little fixes as followup patches...

ldoktor:
runs OK, let me merge this and send the little adjustments...

[job.zip](https://github.com/avocado-framework/avocado-misc-tests/files/401623/job.zip)

PR #75, opened 2016-06-22
Adding sensors command Test

harish-24:

Added a sensors command test for avocado.

### DIFF
--- a/cpu/sensors.py
+++ b/cpu/sensors.py
@@ -1,0 +1,116 @@
+#! /usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harish <harisrir@linux.vnet.ibm.com>
+#
+
+"""
+Test for sensors command
+"""
+import platform
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import distro
+from avocado.utils.software_manager import SoftwareManager
+
+# TODO: Add possible errors of sensors command
+ERRORS = ['I/O error']
+
+
+class Sensors(Test):
+
+    """
+    Run sensors command on ppc architectures.
+    """
+
+    @staticmethod
+    def check_errors(cmd):
+        """
+        Checks any errors in command result
+        """
+        errs = []
+        s_output = process.system_output(cmd)
+        for error in ERRORS:
+            if error in s_output:
+                errs.append(error)
+        return errs
+
+    def test(self):
+        """
+        Test for sensors command
+        """
+        s_mg = SoftwareManager()
+        d_distro = distro.detect()
+        if d_distro.name == "Ubuntu":
+            if not s_mg.check_installed("lm-sensors") and not s_mg.install(
+                    "lm-sensors"):
+                self.error('Need sensors to run the test')
+        else:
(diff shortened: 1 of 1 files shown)


cpu/sensors.py
```diff
@@ -0,0 +1,116 @@
...
+        s_mg = SoftwareManager()
+        d_distro = distro.detect()
+        if d_distro.name == "Ubuntu":
+            if not s_mg.check_installed("lm-sensors") and not s_mg.install(
+                    "lm-sensors"):
+                self.error('Need sensors to run the test')
+        else:
+            if not s_mg.check_installed("lm_sensors") and not s_mg.install(
+                    "lm_sensors"):
+                self.error('Need sensors to run the test')
```
PraveenPenguin, line 64: i think if else condition for "`distro check`" is not required (in code both having same logic)

cpu/sensors.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test(self):
+        """
+        Test for sensors command
+        """
+        s_mg = SoftwareManager()
+        d_distro = distro.detect()
+        if d_distro.name == "Ubuntu":
+            if not s_mg.check_installed("lm-sensors") and not s_mg.install(
+                    "lm-sensors"):
```
PraveenPenguin, line 59: this has to be an `or` instead of an `and`

cpu/sensors.py
```diff
@@ -0,0 +1,116 @@
...
+        error_list = self.check_errors('sensors -f')
+        if len(error_list) > 0:
+            self.error('sensors -f command failed with %s' % error_list)
+        error_list = self.check_errors('sensors -A')
+        if len(error_list) > 0:
+            self.error('sensors -A command failed with %s' % error_list)
+        error_list = self.check_errors('sensors -u')
+        if len(error_list) > 0:
+            self.error('sensors -u command failed with %s' % error_list)
+
```
PraveenPenguin, line 114: can this self.error replaced to self.fail() as self.fail  will prompt as 'TestFail:'

cpu/sensors.py
harish-24, line 64: The package installed is different for each case. `lm-sensors` and `lm_sensors`. Thanks.

cpu/sensors.py
harish-24, line 59: This line makes sure the package is installed. Only if both the conditions fail, error has to be raised. So `or` would invoke a test error without trying to install the package.

cpu/sensors.py
harish-24, line 114: Sure.

cpu/sensors.py
apahim, line 59: @PraveenPenguin, @harish-24 is right. I guess you commented this due to my wrong comment on your perfmon PR, so let's clarify everywhere :)

cpu/sensors.py
apahim, line 64: The repetition of code can be a sign to have a function.  It could receive the package list, install those not already  installed and return true/false on success/error.
But my concern here is the lack of distros. For example, this test will fail on debian if `lm-sensors` is not present.
Also, this check is everywhere on our tests in this repo. Maybe we should make it part of SoftwareManager. I will open a Trello card for Avocado so we can evaluate that.

harish-24, line 64: @apahim Yes. Most tests need some packages to be installed. As mentioned, problem occurs when package name is different across distros..

apahim, line 64: The need of the check is not under discussion. I think we all agree it's needed. The way it's checked is the point. Personally, I like this approach:
https://github.com/avocado-framework/avocado-misc-tests/blob/master/generic/gdb.py#L33-L43
It does not mean you have to change yours, since it's working.
And for an even more clean way to do the dependency check/install (in the future), I created this card:
https://trello.com/c/wpOjFiHl/791-make-the-test-dependencies-installation-part-of-softwaremanager

PraveenPenguin, line 64: @harish-24 thanks pointing in  this , my mistake :(

@apahim  yes this proposal  will help .

apahim:
This repository is frozen. Please send the PR with the requested changes to:
https://github.com/avocado-framework-tests/avocado-misc-tests

PR #76, opened 2016-06-22
Added hackbench test to avocado

PraveenPenguin:
added hackbench Test case

### DIFF
--- a/perf/hackbench.py
+++ b/perf/hackbench.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+# Based on code by Nikhil Rao <ncrao@google.com>
+#   copyright: 2008 Google
+#   https://github.com/autotest/autotest-client-tests/tree/master/hackbench
+
+import os
+import shutil
+import json
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process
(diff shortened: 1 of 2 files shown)


perf/hackbench.py
```diff
@@ -0,0 +1,79 @@
...
+            cc = 'cc'
+        process.system('%s  hackbench.c -o hackbench -lpthread' % cc)
+
+    def test(self):
+
+        num_groups = self.params.get('num_groups', default=90)
+        hackbench_bin = os.path.join(self.srcdir, 'hackbench')
+        cmd = '%s %s' % (hackbench_bin, num_groups)
+        raw_output = process.system_output(cmd, shell=True)
+        self.results = raw_output
```
apahim, line 68: why not `self.results = process.system_output(cmd, shell=True)`?

perf/hackbench.py
```diff
@@ -0,0 +1,79 @@
...
+        perf_json = {}
+        lines = self.results.split('\n')
+        for line in lines:
```
apahim, line 71: why not `for line in self.results.split('\n')`?

apahim:
Is there a condition that could be used to consider the test as `FAIL`? I mean, we are running `hackbench` and collecting the time... what is a bad result? Maybe a threshold in time?

PraveenPenguin:
@apahim  threshold  time of hackbanch differs distro to distro 

as example below ubuntu behaviour as below 

```
   user@host: ~ $ hackbench
   Running in process mode with 10 groups using 40 file  descriptors  each
   (== 400 tasks)
   Each sender will pass 100 messages of 100 bytes
   Time: 0.890
```

so not sure which can be common for all distribution  as well as upto groups so it up to user to after run and analyze the result .It also 

can please help me on this .

Thanks in advance

perf/hackbench.py
PraveenPenguin, line 71: @apahim  thanks will interoperate all the suggestion and raise a new PR.

Thanks

PraveenPenguin:
@apahim  can please help me  threshold time , if you agree what is proposed earlier that user need identify  the failure , will request new PR with changes

Thanks

apahim:
@PraveenPenguin the threshold could then be configurable, defaults to no threshold.

PraveenPenguin:
@apahim  thanks for input , will update code with suggested change and open in new PR

Thanks

PraveenPenguin:
@apahim added new PR #116   with change 

Thanks

PR #53, opened 2016-06-09
Test to verify basic RAS tools commands on Power platform

Pavithra1602:
Python script ras.py executes the scripts listed in mux.yaml which inturn validates RAS tools.
Command to run:
avocado run ras.py --multiplex ras.py.data/mux.yaml

PR #78, opened 2016-06-24
Test to verify basic RAS tools commands on Power platform - v2

Pavithra1602:
Link to v1 - https://github.com/avocado-framework/avocado-misc-tests/pull/53


generic/utils/ChangeLog
```diff
@@ -0,0 +1,10 @@
+1. Added license heading.
+2. Added documentation string.
+3. Fixed variable naming convention errors.
+4. Modified package installation code using for loop.
+5. Modified code to use absolute paths.
+6. Used parenthesis to wrap the line.
+7. Modified code to run command only once using process library.
+8. Removed print statement.
+9. Modified message logged during failure.
+10. Wrote new common_functions.sh script.
```
ldoktor, line 10: Cool, changelog. Although I meant the change log only for pull request description, we don't really need to store it inside utils. I'm sorry I wasn't clear on that. We need to remove this file before the merge. But thank you for it :smile:

generic/utils/ras.py
```diff
@@ -0,0 +1,61 @@
...
+    '''
+    This script executes the scripts listed in mux.yaml which inturn validates RAS tools.
+    Command to run:
+    avocado run ras.py --multiplex ras.py.data/mux.yaml
+    '''
+
+    def setUp(self):
+        if os.geteuid() != 0:
+            exit("You need to have root privileges to run this script."
+                 "\nPlease try again, using 'sudo'. Exiting.")
```
ldoktor, line 36: Well it's not really recommended running avocado as root. There is an API to run executed commands using sudo, thought. So it'd be better to remove this check and use `sudo=True` for commands requiring superuser (see the comment below)

generic/utils/ras.py
```diff
@@ -0,0 +1,61 @@
...
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error("Fail to install %s required for this test." % package)
+
+    def test(self):
+        script_to_run = self.params.get("script", default=None)
+        tool_to_run = self.params.get("tool", default=None)
+        log_dir = os.path.join(self.datadir, 'log')
+        if not os.path.exists(log_dir):
+            os.makedirs(log_dir)
+        command_to_run = (self.datadir + '/' + script_to_run + ' ' +
```
ldoktor, line 51: please use `os.path.join` for joining paths

generic/utils/ras.py
```diff
@@ -0,0 +1,61 @@
...
+                          '>>' + ' ' + self.datadir + '/log/output.txt' + ' ' + '2>&1')
+        cmd_result = process.run(command_to_run, ignore_status=True)
```
ldoktor, line 53: Here you can use `sudo=True`. When the uid is 0 there is no change, but when this is executed under non-privileged user it prepends `sudo` to it.

generic/utils/ras.py
```diff
@@ -0,0 +1,61 @@
...
+        self.log.info(cmd_result)
```
ldoktor, line 54: Well it's not really needed to log the full cmd_result as the `command`, `stdout` and `stderr`. I think you should only log the exit_status in case the test fails.

generic/utils/ras.py.data/common_functions.sh
```diff
@@ -0,0 +1,50 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
```
ldoktor, line 15: In the previous PR this file was authored by @lmr. I can see it differs a bit, but I really don't know how much. So I need to ask you, did you re-implement this from scratch or did you use the previously existing files? You need to use the files accordingly to the license, so if it's compatible you can use it, but you need to give credits to the author. And depending on the license you might not have the permission to modify it or even use it here. Do you have any links to the original code? Was it already upstream somewhere? I only saw Lucas's name in the previous PR but most of the scripts were without an author.

ldoktor, line 15: https://github.com/avocado-framework/avocado-misc-tests/pull/53#discussion_r67103367

Pavithra1602:
Hi,

Raised new Pull request #88 . Closing this PR.

Thanks,
Pavithtra

PR #80, opened 2016-06-28
Adding ethtool tests for Avocado [v5]

narasimhan-v:

### DIFF
--- a/io/net/ethtool/ethtool_tests.sh
+++ b/io/net/ethtool/ethtool_tests.sh
@@ -1,0 +1,67 @@
+#!/bin/bash -e
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+# Tests the network driver and interface with 'ethtool' command.
+# Different parameters are specified in Parameters section of multiplexer file.
+# Interfaces are specified in Interfaces section of multiplexer file.
+# This test needs to be run as root.
+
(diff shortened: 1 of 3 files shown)


io/net/ethtool/ethtool_tests.sh
```diff
@@ -30,7 +30,6 @@ if SoftwareManager().check_installed("ethtool") is False:
         print "Not able to install ethtool"
         exit(1)
 END
-exit
```
apahim: Could you please squash this commit into the other one?

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,67 @@
...
+echo "Parameter: ${arg}"
+
+ifup $interface
+for command in ifdown ifup; do
+    sleep 10
+    $command $interface
+    sleep 10
+    echo "Ethool, $command:"
+    output_file="$AVOCADO_TEST_OUTPUTDIR/output_$command"
+    ethtool ${arg} ${interface} $action_elapse | tee "$output_file"
```
apahim, line 51: This is not redirecting the stderr to the `$output_file`. In consequence, the the `elif` in line 57 will not find the _Operation not supported_ message (the same applies to the other `elif`s).

io/net/ethtool/ethtool_tests.sh
```diff
@@ -0,0 +1,67 @@
...
+    error=${PIPESTATUS[0]}
+
+    if [ $error == 0 ]; then
+        avocado_debug "PASS"
+    elif grep "Operation not supported $output_file"; then
```
apahim, line 56: This is doomed to hang. You're executing: 

```
$ grep "Operation not supported $output_file"
```

So you are grepping a string nowhere. I mean, the `$output_file` is part of the string you're grepping. It should be:

```
$ grep "Operation not supported" $output_file
```

Sorry not noticing this early, but only now I could get to this point.

apahim, line 56: Same applies to the other `grep`s

apahim:
@narasimhan-v did you actually test the `ethtool` non-zero exit conditions? I mean, did you face them all to check if they're working? The **Operation not supported**, the **no stats available**, and so on. Because the way it is now, I cannot even complete the tests. Please verify your script before send the next version.

narasimhan-v:
@apahim sorry for this, have made all necessary changes, tested and created a new PR #98

PR #83, opened 2016-06-29
Adding interbench test to avocado[V2]

PraveenPenguin:
Signed-off-by: praveen@linux.vnet.ibm.com
added interbench Test case 

v1: https://github.com/avocado-framework/avocado-misc-tests/pull/69

Changes:

``` yaml
v2: Fixed the original copyright
v2: Use the latest version 0.31
v2: Use os.path.join() to join paths
v2: Removed the unnecessary option `-m`
v2: Run the test as root using `sudo`
v2: Removed empty spaces from makefile patch
```

PR #84, opened 2016-06-29
Adding interbench test to avocado[V3]

PraveenPenguin:
added interbench Test case 
V2 #83 

Changes:
v3: Use os.path.join() to join paths
v3: Fixed variable names
v3: Removed empty space from makefile patch

### DIFF
--- a/generic/interbench.py
+++ b/generic/interbench.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


ldoktor:
Thank you, it looks good and behaves also good. I needed to force `TMP` as my one does not contain enough space to run this test (my system partition is smaller then my memory) but then the test runs successfully. Merging...

generic/interbench.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    def setUp(self):
+        '''
+        Build interbench
+        Source:
+        http://ck.kolivas.org/apps/interbench/interbench-0.31.tar.bz2
+        '''
+        sm_manager = SoftwareManager()
+        if not sm_manager.check_installed("gcc") and not sm.install("gcc"):
```
ldoktor, line 47: oups, too late, this value should have been `sm_manager`. Never mind, fixed it in https://github.com/avocado-framework/avocado-misc-tests/commit/304be8856086600cf3323fbd8dd8bba3212fd622

PR #86, opened 2016-06-30
Test to run gdb testsuite.

Pavithra1602:
This python script downloads gdb tar file from http://ftp.gnu.org/gnu/gdb/ and runs gdb testsuite. If number of unexpected failures are greater than zero, result will be marked as FAIL. gdb version can be given in version.yaml file, by default version will be 7.10.

### DIFF
--- a/generic/gdb.py
+++ b/generic/gdb.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+import re
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import build
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class GDB(Test):
+
+    def setUp(self):
+        sm = SoftwareManager()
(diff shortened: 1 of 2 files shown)


apahim:
@Pavithra1602, what distro are you using? Fedora package names are different from the names you provided...

apahim:
I guess it's ubuntu... or debian, right? So the test should be improved with something like this:

``` diff
$ git diff
diff --git a/generic/gdb.py b/generic/gdb.py
index 690ddcf..31b8530 100755
--- a/generic/gdb.py
+++ b/generic/gdb.py
@@ -20,6 +20,7 @@ from avocado import Test
 from avocado import main
 from avocado.utils import archive
 from avocado.utils import build
 from avocado.utils.software_manager import SoftwareManager

@@ -27,11 +28,35 @@ from avocado.utils.software_manager import SoftwareManager
 class GDB(Test):

     def setUp(self):
-        sm = SoftwareManager()
-        for package in ("gcc", "g++", "dejagnu", "flex", "binutils-dev", "bison"):
-            if not sm.check_installed(package) and not sm.install(package):
-                self.error(
-                    "Fail to install %s required for this test." % package)
+        packages = []
+        if dist.name == 'ubuntu':
+            packages = ['gcc',
+                        'g++',
+                        'dejagnu',
+                        'flex',
+                        'binutils-dev',
+                        'bison']
+        elif dist.name == 'fedora':
+            packages = ['gcc',
+                        'gcc-c++',
+                        'dejagnu',
+                        'flex',
+                        'binutils-devel',
+                        'bison']
+        else:
+            self.info("I don't have the package list for your distro. "
+                      "Let's try it anyway.")
+            # Or, if you prefer to fail here...
+            # self.fail('no packages list for your distro.')
+
+        if packages:
+            sm = SoftwareManager()
+            for package in packages:
+                if not sm.check_installed(package) and not sm.install(package):
+                    self.error("Fail to install %s required for this test." %
+                               package)
+
         gdb_version = self.params.get('gdb_version', default='7.10')
         tarball = self.fetch_asset(
             "http://ftp.gnu.org/gnu/gdb/gdb-%s.tar.gz" % gdb_version)

```

Pavithra1602:
Raised new pull request https://github.com/avocado-framework/avocado-misc-tests/pull/108
closing the PR.

PR #88, opened 2016-06-30
Test to verify basic RAS tools commands on Power platform - v3

Pavithra1602:
Link to v2 - https://github.com/avocado-framework/avocado-misc-tests/pull/78

Changelog:
1. Removed check for root user added sudo=True.
2. Modified code to use os.path.join for joining paths.
3. Re-implemented common_functions.sh [now run_cmd_func.sh] and sys_ident_tool.sh which were taken from different source as they were not GPL compatible.


generic/utils/ras.py
```diff
@@ -0,0 +1,57 @@
...
+        architecture = commands.getoutput("arch")
+        if "ppc" not in architecture:
+            self.skip("supported only on Power platform")
+        sm = SoftwareManager()
+        for package in ("ppc64-diag", "powerpc-utils", "lsvpd"):
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error("Fail to install %s required for this test." % package)
+
+    def test(self):
+        script_to_run = self.params.get("script", default=None)
```
ldoktor, line 43: You should specify some meaningful default. None fails critically.

generic/utils/ras.py
```diff
@@ -0,0 +1,57 @@
...
+        tool_to_run = self.params.get("tool", default=None)
+        log_dir = os.path.join(self.datadir, 'log')
+        if not os.path.exists(log_dir):
+            os.makedirs(log_dir)
+        command_to_run = (os.path.join(self.datadir, script_to_run) + ' ' +
```
ldoktor, line 48: you use relative paths to source the other scripts, therefor you have to change the directory to the base location, which currently would be the `os.path.dirname(self.datadir)`. This is a bit odd, so I'd recommend changing the scripts to source from `./` instead of `ras.py.data/` and I'd use the `os.chdir(self.datadir)` to make sure it's executed from the right location.

generic/utils/ras.py
```diff
@@ -0,0 +1,57 @@
...
+                          '>>' + ' ' + os.path.join(self.datadir, "log", "output.txt") + ' ' + '2>&1')
```
ldoktor, line 49: All in all it might look like:

``` python
os.chdir(self.datadir)
script_to_run = self.params.get("script", default="sys_ident_tool.sh")
tool_to_run = self.params.get("tool", default="sys_ident")
log = os.path.join(self.outputdir, "output.txt")
cmd_result = process.run("%s >> %s 2>&1" % (script_to_run, log), ignore_status=True, sudo=True)
```

Which works fine if you remove the `ras.py.data/` prefix in all the scripts.

generic/utils/ras.py
```diff
@@ -0,0 +1,57 @@
...
+            self.skip("supported only on Power platform")
+        sm = SoftwareManager()
+        for package in ("ppc64-diag", "powerpc-utils", "lsvpd"):
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error("Fail to install %s required for this test." % package)
+
+    def test(self):
+        script_to_run = self.params.get("script", default=None)
+        tool_to_run = self.params.get("tool", default=None)
+        log_dir = os.path.join(self.datadir, 'log')
```
ldoktor, line 45: you must never write to `self.datadir`. Please use `self.outputdir` instead.

generic/utils/ras.py
```diff
@@ -0,0 +1,57 @@
...
+class RASTools(Test):
+
+    '''
+    This script executes the scripts listed in mux.yaml which inturn validates RAS tools.
+    Command to run:
+    avocado run ras.py --multiplex ras.py.data/mux.yaml
+    '''
+
+    def setUp(self):
+        architecture = commands.getoutput("arch")
```
ldoktor, line 34: no need to execute any commands, you can use `os.uname()[4]`

lmr:
My suggestion would be to port the simple shell scripts to python. I took a quick look and it seems the conversion should be fairly straight forward.

Pavithra1602:
Raised new pull request https://github.com/avocado-framework/avocado-misc-tests/pull/109
Closing the pull request.

PR #95, opened 2016-07-09
Adding syscall fuzzer(trinity) test to avocado

PraveenPenguin:
added trinity (new) Test case 

```

v1: fetch syscall fuzzer from git (https://github.com/kernelslacker/trinity)
v1: crate a non root user to run test suite 
v1: run ./configure and make
v1: invoke trinity exe 
v1:verify after run if there problem reported in dmesg 
v1: problem reported than notify user to evaluate the problem 
```

### DIFF
--- a/fuzz/trinity.py
+++ b/fuzz/trinity.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+
+import os
+import re
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive, build, process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Trinity(Test):
+
+    """
+    This testsuite test  syscall by calling syscall
+    with random system call and varying number of args
+    """
+
+    def setUp(self):
+        '''
+        Build Trinity
+        Source:
+        https://github.com/kernelslacker/trinity
+        '''
+
+        process.run('useradd  trinity', sudo=True)
+
+        smm = SoftwareManager()
+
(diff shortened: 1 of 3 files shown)


fuzz/trinity.py
```diff
@@ -0,0 +1,88 @@
...
+        smm = SoftwareManager()
+
+        for package in ("gcc", "make"):
+            if not smm.check_installed(package) and not smm.install(package):
+                self.error(
+                    "Fail to install %s required for this test." % package)
+
+        locations = ["https://github.com/kernelslacker/trinity/archive/"
+                     "master.zip"]
+        tarball = self.fetch_asset("trinity.zip", locations=locations)
```
apahim, line 52: Depending on the version of avocado (>=38), you can set the expire time for the asset file, so avocado will download it again if expired. Useful to refresh the file on your cache directory:

```
tarball = self.fetch_asset("trinity.zip", locations=locations, expire='7d')
```

fuzz/trinity.py.data/README
```diff
@@ -0,0 +1,14 @@
...
+   local network.
+
+* Trinity may generate the right selection of syscalls to start
+  sending random network  packets to other hosts. While every effort
+   is made to restrict this to IP addresses  on local lans, multicast
+   & broadcast, care should be taken to not allow the  packets it generates
+   to go out onto the internet.
+
+there is a real good chance that you could screw up your computer. YOU
+HAVE BEEN WARNED!
```
apahim, line 14: Ok, good to know. Running on a vm then :)

fuzz/trinity.py
PraveenPenguin, line 52: @apahim  thanks for looking into this , will create a new pull request with this change 

Thanks

apahim:
I'm having issues to run this test. I'm running it as root, but it fails with permissions denied. The `self.srcdir` is created in `/var/tmp/avocado_<random>/<test>/src/`. When avocado tries to execute `process.system('su - trinity -c " %s  %s  %s...` it fails the test:

```
[stderr] -bash: /var/tmp/avocado_lgONVG/1-fuzz_trinity.py_Trinity.test/src/trinity-master/trinity: Permission denied
```

It happens because `/var/tmp/avocado_lgONVG/` is created with `700`:

```
# ls -ld /var/tmp/avocado_lgONVG/
drwx------. 3 root root 44 Jul 20 09:20 /var/tmp/avocado_lgONVG/
```

So user `trinity` can not execute files inside it. I see you change permissions in line 58, but that not seems to be enough. At least in Fedora.

PraveenPenguin:
@apahim  will incorporate all the comment   and open new PR with fix 

Thanks

PraveenPenguin:
@apahim made changes and request a PR #107 

Thanks

PR #96, opened 2016-07-10
Added Oprofile test 

poojabs7:
Added Oprofile test which covers oprofile test scenarios for opcontrol and operf

### DIFF
--- a/generic/oprofile.py
+++ b/generic/oprofile.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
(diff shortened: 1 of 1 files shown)


generic/oprofile.py
```diff
@@ -0,0 +1,63 @@
...
+            if not (os.path.exists("/usr/bin/oprofiled") and
+                    ("/usr/bin/runtest")):
+                self.skip("Need oprofile and dejagnu packages to run test")
+        git.get_repo('git://oprofile.git.sourceforge.net/gitroot/oprofile/'
+                     'oprofile-tests', destination_dir=self.srcdir)
+        os.chdir(self.srcdir)
+        os.chdir("testsuite/")
+
+    def test(self):
+        result = process.system('runtest --tool oprofile', ignore_status=True,
```
apahim, line 55: This `result` is not used. You can drop it if you prefer.

generic/oprofile.py
```diff
@@ -0,0 +1,63 @@
...
+                                sudo=True)
+        pattern = re.compile("FAIL")
+        for i, line in enumerate(open('oprofile.sum')):
```
apahim, line 58: It took some time to understand that you are looking for the lines with the string `FAIL` in it. There are simpler ways to do that, but it's working.
What do you think about to show the lines which contains the `FAIL` in it? That way user can see the actual error, given the `oprofile.sum` file will disappear  after the test.

apahim:
Ping @poojabs7. Could you send a v2 with the requested improvements? Closing this for now doe to the lack of activity.

PR #99, opened 2016-07-14
Added centos distribution to the configuration

sathnaga:
Added centos distribution to the test configuration


apahim:
Ok, thanks.

PR #100, opened 2016-07-15
Adding scsi remove and add test for Avocado [v3]

narasimhan-v:

### DIFF
--- a/io/disk/scsi_remove_add.sh
+++ b/io/disk/scsi_remove_add.sh
@@ -1,0 +1,64 @@
+#!/bin/bash -e
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+# This Test removes and adds back a scsi device in all the specified PCI 
+# domains specified in the 'multiplexer' file.
+# Runs the test for '0000:01:00.0' if no value for pci domain is given.
+# This test needs to be run as root. 
+
+PATH=$(avocado "exec-path"):$PATH
+
+# Install dependencies
+python << END
+from avocado.utils.software_manager import SoftwareManager;
+if SoftwareManager().check_installed("lsscsi") is False:
+    print "lsscsi is not installed"
+    if SoftwareManager().install("lsscsi") is False:
+        print "Not able to install lsscsi"
+        exit(1)
+END
+
+[[ -z $pci_device ]] && pci_device="0000:01:00.0" 
+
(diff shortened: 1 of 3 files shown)


narasimhan-v:
Attaching `job.log`
[job_log.txt](https://github.com/avocado-framework/avocado-misc-tests/files/365289/job_log.txt)

io/disk/scsi_remove_add.sh
```diff
@@ -0,0 +1,64 @@
...
+        exit(1)
+END
+
+[[ -z $pci_device ]] && pci_device="0000:01:00.0" 
+
+echo "PCI Device: $pci_device"
+device_list_output=($(ls -l /dev/disk/by-path/ | grep "$pci_device" | awk '{print $NF}'))
+if [[ -z $device_list_output ]]; then
+    avocado_debug "No Devices in PCI ID $pci_device"
+    exit
```
apahim, line 40: Is this a good ending? Avocado is showing `PASS` as test result when one enters this `if`.

io/disk/scsi_remove_add.sh
```diff
@@ -0,0 +1,64 @@
...
+    sleep 5
+    lsscsi
+    echo "$scsi_num deleted"
+    echo "adding $scsi_num back"
+    echo "scsi add-single-device $scsi_num_seperated" > /proc/scsi/scsi
+    sleep 5
+    lsscsi
+    echo "$scsi_num added back"
+    echo
+done
```
apahim, line 64: When will the test actually fail? O don't see an `exit 1` anywhere...

io/disk/scsi_remove_add.sh
narasimhan-v, line 40: We do not have a way to skip tests in bash scripts.
Can you suggest a better way to handle this ?

io/disk/scsi_remove_add.sh
narasimhan-v, line 64: Test will fail when any of the commands in line 54 or line 59 fails.

io/disk/scsi_remove_add.sh
apahim, line 40: `exit 1`

io/disk/scsi_remove_add.sh
apahim, line 64: Ok, sorry, I didn't notice the `-e` in the shebang.

io/disk/scsi_remove_add.sh
narasimhan-v, line 40: So assumption is, instead of `SKIP` we can `FAIL` ?

apahim, line 40: Yes, we don't have skip. But it's your call... I'm just wondering what would be a better fit here (`PASS` or `FAIL`). I'm ok with both.

narasimhan-v, line 40: I did not want it to `FAIL`. So, `PASS` is ok for me, till a way to skip is introduced for bash scripts.

apahim, line 40: Ok, nice.

apahim:
LGTM

io/disk/scsi_remove_add.sh
narasimhan-v, line 40: I would prefer a `PASS` over a `FAIL`

narasimhan-v, line 40: @apahim Thanks

PR #105, opened 2016-07-20
Adding xfstests test to avocado

PraveenPenguin:
added xfstests Test case

```

v1: fetch xfstest suite from git (git://oss.sgi.com/xfs/cmds/xfstests)
v1: add README File (for requirement as well  test ) 
v1: run ./configure and make
v1: add test user 
v1: add function to list all test 
v1:add group file to run file system suite use group
v1: verify test run and inform user test run status
```

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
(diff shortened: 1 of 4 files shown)


apahim:
Could you please provide the job log? Thank you.

PraveenPenguin:
@apahim please find job log in below link

https://paste.fedoraproject.org/401681/04011981/

Thanks

fs/xfstests.py
```diff
@@ -0,0 +1,168 @@
...
+        skip_dangerous = self.params.get('skip_dangerous', default=True)
+        if test_number == '000':
+            self.log.info('Dummy test to setup xfstests')
+            return
+
+        if test_number not in self._get_available_tests():
+            raise self.error('test file %s not found' % test_number)
+
+        if skip_dangerous:
+            if test_number in self._get_tests_for_group('dangerous'):
```
apahim, line 156: Overall, this test looks good. I would move this 'skip_dangerous'  part to `setUp()`, since there we can use `self.skip()`, which is a perfect fit for this case. Notice that `self.skip()` cannot be used outside `setUp()`. Since `setUp()` is executed once per test, you can do all the parametrization there, and skip a test when applies. Please let me know if you are going to make such improvement or if you prefer to keep the test as is.

PraveenPenguin, line 156: @apahim yes that make sense , thanks will make changes and raise new PR

PraveenPenguin:
@apahim adopt new changes and raise PR #143, please have a look

Thanks 
Praveen

PR #106, opened 2016-07-20
Adding blkdiscard test for SSD devices [v3]

vrbagalkote:
Signed-off-by: Venkat R B vrbagal1@linux.vnet.ibm.com

### DIFF
--- a/io/disk/ssd/blkdiscard.py
+++ b/io/disk/ssd/blkdiscard.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2016 IBM
+# Author: Venkat Rao B <vrbagal1@linux.vnet.ibm.com>
+
+# Blkdiscard  is  used  to  discard  device  sectors.This is useful for
+# solid-state drivers (SSDs) and thinly-provisioned storage.
+
+from avocado import Test
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import process
+from avocado import main
+
+
+class Blkdiscard(Test):
+
+    """
+    blkdiscard is used to discard device sectors.This is useful for
+    solid-state drivers (SSDs) and thinly-provisioned storage.
+    """
+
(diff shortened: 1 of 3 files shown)


io/disk/ssd/blkdiscard.py
```diff
@@ -0,0 +1,76 @@
...
+                    cmd = "blkdiscard %s -o %d -l %d -v" \
+                        % (self.disk, int(size) / i, int(size) / j)
+                    process.system(cmd, shell=True)
+                else:
+                    cmd = "blkdiscard %s -o %d -l %d -v" \
+                        % (self.disk, int(size) / i, int(size) / j)
+                    if process.system(cmd, ignore_status=True,
+                                      shell=True) == 0:
+                        self.fail("Blkdiscard passed for the values which is, \
+                            not aligned to 4096 but actually it should fail")
```
apahim, line 73: Wondering... How about to create one `test_xxxx()` function for each test you have?  Like:

``` ptyhon
...
    def test_foo(self):
        cmd = "blkdiscard %s -o 0 -v -l %s" % (self.disk, self.size)
        process.run(cmd, shell=True)

    def test_bar(self):
        cmd = "blkdiscard %s -o %s -v -l %s" % (self.disk, self.size, self.size)
        process.run(cmd, shell=True)
...
    def test_bigone(self):
        for i in xrange(2, 10, 2):
        ...
```

With `self.size` being created in `setUp()`.

What do you think? That way user could follow-up in the TUI what test is running and so on...

vrbagalkote, line 73: Basically, scripts runs the same test case, for different values. Its not a different test case. So I think its appropriate, to have the test case under one function.

apahim, line 73: ok. I will test it asap.

apahim:
@vrbagalkote could you provide the job.log file for this test? You can use http://fpaste.org or something like that.

vrbagalkote:
@apahim Please find the job.log o/p in the below link.

https://paste.fedoraproject.org/393878/91903901/

apahim:
@vrbagalkote thank you for the file. Isn't this an error?

```
2016-07-22 08:18:43,779 process          L0420 DEBUG| [stderr] blkdiscard: /dev/nvme0n1: offset 400078973952 is not aligned to sector size 4096
```

I mean, shouldn't it make the avocado test to `FAIL` instead of `PASS`?

vrbagalkote:
Yes, I am checking for the command exit status, if its fails then it is as expected. In other words, its a part of negative testing. If the command passes even for the values that are not aligned to the disk block size, then the script will fail.

vrbagalkote:
if process.system(cmd, ignore_status=True,
                                     shell=True) == 0:
    self.fail("Blkdiscard passed for the values which is, \
                             not aligned to 4096 but actually it should fail")
@apahim

apahim:
Ok, got it. Thank you.

PR #108, opened 2016-07-21
Test to run gdb testsuite - v2

Pavithra1602:
link to v1: https://github.com/avocado-framework/avocado-misc-tests/pull/86

As per comment added check for distro.

### DIFF
--- a/generic/gdb.py
+++ b/generic/gdb.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+import re
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import build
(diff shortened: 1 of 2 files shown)


generic/gdb.py
```diff
@@ -0,0 +1,77 @@
...
+            packages = ['gcc',
+                        'gcc-c++',
+                        'dejagnu',
+                        'flex',
+                        'binutils-devel',
+                        'bison']
+        else:
+            self.fail('no packages list for your distro.')
+
+        if packages:
```
apahim, line 51: This `if` is meaningless since you're going to fail in line 51.

generic/gdb.py
```diff
@@ -0,0 +1,77 @@
...
+        dist = distro.detect()
+        packages = []
+        if dist.name == 'Ubuntu':
+            packages = ['gcc',
+                        'g++',
+                        'dejagnu',
+                        'flex',
+                        'binutils-dev',
+                        'bison']
+        elif dist.name == 'SuSE' or dist.name == 'redhat' or dist.name == 'fedora':
```
apahim, line 41: nice

apahim:
Ok, testing.

generic/gdb.py
clebergnu, line 41: Style suggestion: `elif dist.name in ['SuSE', 'redhat', 'fedora']:`

apahim, line 41: Yes, good one. And since we are talking about styling, as (I suspect) you're not going to test `if packages` in line 51, you could avoid code duplication by appending only the packages that are different:

``` python
...
packages = ['gcc', 'dejagnu', 'flex', 'bison']
if dist.name == 'Ubuntu':
    packages.extend(['g++', 'binutils-dev'])
...
```

apahim:
Tests are ok. Will wait your answer on the minor comments we have. Thank you.

Pavithra1602:
Hi,

Raised new pull request #113. closing this pull request.

Thanks,
Pavithra

PR #109, opened 2016-07-21
Ras tools v4

Pavithra1602:
link to v3 - https://github.com/avocado-framework/avocado-misc-tests/pull/88
## ChangeLog
1. Added all the tools tests as functions instead of running separate shell scripts.
2. Modified command for checking architecture


generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+    nvram - nvram command retrieves and displays NVRAM data
+    ofpathname - ofpathname translates the device name between logical name and Open Firmware name
+    ppc64_cpu - ppc64_cpu is used to set cpu options
+    rtas_ibm_get_vpd - rtas_ibm_get_vpd gives vpd data
+    rtas_errd and rtas_dump - rtas_errd adds RTAS events to /var/log/platform and rtas_dump dumps RTAS events
+    '''
+    is_fail = 0
+
+    def run_cmd(self, cmd):
+        global is_fail
```
ldoktor, line 46: Well this is actually not a global variable, it's an attribute of `RASTools` test and you already use it as `self.is_fail` so you can simply remove this...

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+        print "Output of command \"%s\":" % cmd
```
ldoktor, line 47: You should always use `self.log` instead of print. Anyway you can also omit it as `process.run` logs the executed command.

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+        cmd_result = process.run(cmd, ignore_status=True, sudo=True)
+        if cmd_result.exit_status != 0:
+            self.is_fail = 1
```
ldoktor, line 50: How about `self.is_fail += 1` and then in tests you can even say that there were `self.is_fail` soft-errors during the execution? (don't take this mandatory, though)

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+        return
+
+    def setUp(self):
+        architecture = os.uname()[4]
+        if "ppc" not in architecture:
+            self.skip("supported only on Power platform")
+        sm = SoftwareManager()
+        for package in ("ppc64-diag", "powerpc-utils", "lsvpd"):
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(
```
ldoktor, line 60: This is valid, but pretty ugly formatting. Would you please use something like this in case it does not fits properly:

``` python
self.error("Fail to install %s"
           "required for this"
           "test."
           % package)
```

that looks much nicer...

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+                    "Fail to install %s required for this test." % package)
+        log_dir = os.path.join(self.outputdir, 'log')
+        if not os.path.exists(log_dir):
+            os.makedirs(log_dir)
+
+    def test1_set_poweron_time(self):
+        print "===============Executing set_poweron_time tool test==============="
```
ldoktor, line 67: Please don't use prints, use `self.log` instead (for these lines I'd even use `self.log.info` as it's quite important line)

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+        self.run_cmd("set_poweron_time -m")
+        self.run_cmd("set_poweron_time -h")
+        self.run_cmd("set_poweron_time -d m2")
+        self.run_cmd("set_poweron_time -t M6D15h12")
+        if self.is_fail == 1:
+            self.fail("RAS check failed")
```
ldoktor, line 73: If you decide to use my recommendation, this would have to become:

``` python
if self.is_fail:
    self.fail("%s RAS check(s) failed" % self.is_fail)
```

I think that would be a useful info :wink:

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def test3_lsmcode(self):
+        print "===============Executing lsmcode tool test==============="
+        self.run_cmd("lsmcode")
+        self.run_cmd("lsmcode -A")
+        self.run_cmd("lsmcode -v")
+        self.run_cmd("lsmcode -D")
+        path_db = commands.getoutput("sudo find / -iname vpd.db | head -1")
```
ldoktor, line 88: Please use `process.system_output(..., sudo=True, shell=True)` instead of running `commands` directly.

ldoktor, line 88: Btw can I ask what is the purpose of this? I don't know the `vpd.db` but scanning the whole machine to try to find any matching file seems really dangerous.

Also the `head -1` would likely report the first error message so you should use `2>/dev/null` in the `find` command, but please reconsider the whole find...

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+        if path_db:
+            copyfile_path = os.path.join(self.outputdir, 'vpd.db')
+            copyfile(path_db, copyfile_path)
+            self.run_cmd("lsmcode --path=%s" % copyfile_path)
+        path_tar = commands.getoutput("sudo find / -iname vpd.*.gz | head -1")
```
ldoktor, line 93: Oh this is even worse. What is this file? Can't it be packaged in the test's `.data` dir? Or isn't there a better way of finding it?

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+import commands
```
ldoktor, line 18: You should never need the `commands` import as avocado wraps it inside the `avocado.process`, which is what you should be using. There are additional benefits in the `process` library which you are giving away by not using it.

ldoktor, line 18: The same applies for `print` you should always use `self.log` or `import logging; logging.getLogger("avocado.test")` inside tests.

ldoktor:
Hello @Pavithra1602, I have to say I'm amazed. I hesitated to ask you to rewrite it in python, but it just looks **way** nicer. Thank you for all the work. Please take a look at the few comments, there is one suggestion and few repeated little mistakes like using `prints`, `commands` and such.

There is one think I don't understand and that's the `sudo find / ...`. Can you please provide some info regarding that? Searching the **whole** machine just to locate one file is really ugly.

generic/utils/ras.py
```diff
@@ -0,0 +1,229 @@
...
+    def setUp(self):
+        architecture = os.uname()[4]
+        if "ppc" not in architecture:
+            self.skip("supported only on Power platform")
+        sm = SoftwareManager()
+        for package in ("ppc64-diag", "powerpc-utils", "lsvpd"):
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(
+                    "Fail to install %s required for this test." % package)
+        log_dir = os.path.join(self.outputdir, 'log')
```
ldoktor, line 62: The `self.outputdir` is already an empty directory for any test data. I'm not sure why do you want to add `log` directory inside. I mean it's not a problem, you can do that, I just find it unnecessary. But if you're used to it and you like the results, why not...

Pavithra1602:
Raised new PR #115.
Closing the pull request.

Thanks,
Pavithra

PR #107, opened 2016-07-21
Adding trinity test to avocado[v2]

PraveenPenguin:
Signed-off-by: praveen@linux.vnet.ibm.com
added trinity Test case
V1: #95 

```
v2:set the expire time for the asset file
v2:fix user permission to run script from non - root user
```

PR #112, opened 2016-07-25
Adding trinity test to avocado[v3]

PraveenPenguin:
added trinity Test case
v1 #95 
v2 #107 

`v3:fix issue with Fedora vm call all system calls`

### DIFF
--- a/fuzz/trinity.py.data/trinity.yaml
+++ b/fuzz/trinity.py.data/trinity.yaml
@@ -1,0 +1,36 @@
+#./trinity
+# --arch, -a: selects syscalls for the specified architecture (32 or 64). Both by default.
+# --bdev, -b <node>:  Add /dev node to list of block devices to use for destructive tests..
+# --children,-C: specify number of child processes
+# --debug,-D: enable debug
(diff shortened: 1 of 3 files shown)


fuzz/trinity.py.data/trinity.yaml
```diff
@@ -0,0 +1,36 @@
...
+# --victims,-V: path to victim files.
+
+# -c#,@: target specific syscall (takes syscall name as parameter and optionally 32 or 64 as bit-width. Default:both).
+# -N#: do # syscalls then exit.
+# -s#: use # as random seed.
+
+setup:
+ trinityrun : !mux
+  exclude:
+   runargs :  -x madvise
```
apahim, line 32: Wondering if that `madasive` we were facing issues with wasn't a typo of `madvise`, which is an actual syscall (http://man7.org/linux/man-pages/man2/madvise.2.html)?

PraveenPenguin:
@apahim oh yes you are right , but now test can run all the available system call can you please merge this .

Thanks in advance !!

apahim:
Looks good. Thank you.

PR #113, opened 2016-07-28
Test to run gdb testsuite - v3

Pavithra1602:
link to v2 - https://github.com/avocado-framework/avocado-misc-tests/pull/108

Change log:
1. Removed check for packages.
2. Modified coding style in assigning packages list.

### DIFF
--- a/generic/gdb.py
+++ b/generic/gdb.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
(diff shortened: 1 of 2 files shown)


apahim:
Checking.

generic/gdb.py
```diff
@@ -0,0 +1,64 @@
...
+        self.srcdir = os.path.join(
+            self.srcdir, os.path.basename(tarball.split('.tar')[0]))
+        os.chdir(self.srcdir)
+        process.run('./configure', ignore_status=True, sudo=True)
+        build.make(self.srcdir)
+
+    def test(self):
+        process.run("make check-gdb", ignore_status=True, sudo=True)
+        logfile = os.path.join(self.logdir, "stdout")
+        fp = open(logfile, "r")
```
apahim, line 57: This is keeping the file open and then avocado cannot remove the srcdir after the test:
It is a best practice to use `with` to deal with file objects, so the file is always closed after the commands inside the with....

```
with open(file, 'r') as f:
    for line in f.readlines():
...
```

Pavithra1602:
Raised new PR #114.
Closing this Pull request.

Thanks,
Pavithra

generic/gdb.py
```diff
@@ -0,0 +1,64 @@
...
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+import re
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import build
+from avocado.utils import distro
+from avocado.utils import process
```
PraveenPenguin, line 24: can not be do like this way 
from avocado.utils import distro, archive, process, build

PR #114, opened 2016-07-29
Test to run gdb testsuite - v4

Pavithra1602:
link to v3 - https://github.com/avocado-framework/avocado-misc-tests/pull/113

### DIFF
--- a/generic/gdb.py
+++ b/generic/gdb.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


generic/gdb.py
```diff
@@ -0,0 +1,64 @@
...
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+import re
+from avocado import Test
+from avocado import main
+from avocado.utils import archive
+from avocado.utils import build
+from avocado.utils import distro
+from avocado.utils import process
```
PraveenPenguin, line 24: can not be do like this way
from avocado.utils import distro, archive, process, build

PraveenPenguin, line 24: otherwise looks good to me

apahim, line 24: According to pep8, both are ok: https://www.python.org/dev/peps/pep-0008/#imports

apahim:
Thank you!

PR #115, opened 2016-07-29
Test to verify basic RAS tools commands on Power platform - v5

Pavithra1602:
link for v4 - https://github.com/avocado-framework/avocado-misc-tests/pull/109

Change Log:
1. Removed "global is_fail".
2. Used self.log instead of print.
3. Modified code to log number of failed commands.
4. Modified find command to search only in /var/lib/lsvpd/ directory.
5. Modified code to use self.outputdir directly instead of creating log directory inside.
6. Facing issue in using process.system_output(..., sudo=True) retaining commands.getoutput
[stderr]     raise CmdError(cmd, sp.result)
[stderr] avocado.utils.process.CmdError: Command 'find /var/lib/lsvpd/ -iname vpd.db | head -1' failed (rc=1)
Exit status: 1


ldoktor:
Hello @Pavithra1602 it looks almost good. The `find` location is much better now and all the output dirs seems to be adjusted properly. There are few style fixes and the getoutput issues, please take a look at my proposed fixes https://github.com/avocado-framework/avocado-misc-tests/pull/119 and let me know if that works for you (or suggest changes)

PR #116, opened 2016-08-01
Adding hackbench test to avocado[V2]

PraveenPenguin:
added hackbench Test case
v1 #76 

```
V2: removed extra  variables 
V2: set threshold time in yaml file 
```

### DIFF
--- a/perf/hackbench.py
+++ b/perf/hackbench.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+# Based on code by Nikhil Rao <ncrao@google.com>
+#   copyright: 2008 Google
+#   https://github.com/autotest/autotest-client-tests/tree/master/hackbench
+
+import os
+import shutil
+import json
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Hackbench(Test):
+
+    """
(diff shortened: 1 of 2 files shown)


perf/hackbench.py
```diff
@@ -0,0 +1,84 @@
...
+        if 'CC' in os.environ:
+            cc = '$CC'
+        else:
+            cc = 'cc'
+        process.system('%s  hackbench.c -o hackbench -lpthread' % cc)
+
+    def test(self):
+
+        time_value = self.params.get('time_val', default=None)
+        num_groups = self.params.get('num_groups', default=10)
```
apahim, line 65: If you do the `self.params.get()` on `setUp()`, the parameters will be available for all tests. Ok, you have only one here, but still, it's a good practice.

perf/hackbench.py
```diff
@@ -0,0 +1,84 @@
...
+        hackbench_bin = os.path.join(self.srcdir, 'hackbench')
+        cmd = '%s %s' % (hackbench_bin, num_groups)
+        self.results = process.system_output(cmd, shell=True)
+        perf_json = {}
+        for line in self.results.split('\n'):
+            if line.startswith('Time:'):
+                time_val = line.split()[1]
+                time = time_val
```
apahim, line 73: What about `time = line.split()[1]`?
Also, this `time`/`time_val`/`time_value` is very confusing. I'd call:
`time_limit = self.params.get('time_limit', default=None)`
and
`time_spent = line.split()[1]`
And use only those two terms.

perf/hackbench.py
```diff
@@ -0,0 +1,84 @@
...
+                perf_json = {'time': time_val}
+        output_path = os.path.join(self.outputdir, "perf.json")
+        json.dump(perf_json, open(output_path, "w"))
+        self.log.info("Time Taken:" + time)
+        if time_value:
+            if time_value != time:
```
apahim, line 79: I don't get this. Shouldn't the check be `if time >= time_value:` ?

perf/hackbench.py
PraveenPenguin, line 73: thanks for looking in to this , sure that make sense will do

perf/hackbench.py
PraveenPenguin, line 79: @apahim sure that make sense

Thanks

perf/hackbench.py
PraveenPenguin, line 65: @apahim 
thanks for pointing on this , sure that make sense

PraveenPenguin:
@apahim added new PR #117  with change please have a look

Thanks

PR #117, opened 2016-08-01
Adding hackbench test to avocado[V3]

PraveenPenguin:
added hackbench Test case
v1 #76
v2 #116 

```
v3:rectified variable names 
v3:remove redundant variable
v3: change compare condition
```

### DIFF
--- a/perf/hackbench.py
+++ b/perf/hackbench.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+# Based on code by Nikhil Rao <ncrao@google.com>
+#   copyright: 2008 Google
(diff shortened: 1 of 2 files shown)


perf/hackbench.py
```diff
@@ -0,0 +1,83 @@
...
+        perf_json = {}
+        for line in self.results.split('\n'):
+            if line.startswith('Time:'):
+                time_spent = line.split()[1]
+                perf_json = {'time': time_spent}
+        output_path = os.path.join(self.outputdir, "perf.json")
+        json.dump(perf_json, open(output_path, "w"))
+        self.log.info("Time Taken:" + time_spent)
+        if self._threshold_time:
+            if self._threshold_time >= time_spent:
```
apahim, line 78: Still failing to understand this. Shouldn't it be opposite? I mean, the test fail on `time_spent >= threshold`?

perf/hackbench.py
```diff
@@ -0,0 +1,83 @@
...
+                self.error("Test failed: Time Taken not "
+                           "matched with threshold")
```
apahim, line 80: Message needs update.

perf/hackbench.py
PraveenPenguin, line 78: @apahim  not sure may be what default time i given in yaml file that is not suitable  for your environment , i think user need to edit every time yaml file with threshold time . 
what you think can we remove this check and just go with default run and user need to check after run as i suggested earlier ?

apahim, line 78: Sure. Your test, you decide. The threshold was only a suggestion.

PraveenPenguin, line 78: thanks , will try  once with x86 and power system  with all distribution and try to get higher threshold if there are not major gap , if not will remove this check .

Regards
Praveen

PraveenPenguin:
@apahim  pushed changes in new PR #122 , please have a look.

Thanks

PR #72, opened 2016-06-20
Added connectathon test to avocado

PraveenPenguin:
Signed-off-by: praveen@linux.vnet.ibm.com
added connectathon Test case

PR #129, opened 2016-08-05
Adding connectathon test to avocado[v2]

PraveenPenguin:
Signed-off-by: Praveen K Pandey praveen@linux.vnet.ibm.com
added connectathon Test case

v1: #72 

```
v2: fixed repetition of package check
v2: change test fail check , self.error() to self.fail()
```

### DIFF
--- a/generic/connectathon.py
+++ b/generic/connectathon.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


generic/connectathon.py
```diff
@@ -0,0 +1,101 @@
...
+        if os.geteuid() != 0:
+            exit("You need to have root privileges to run this script."
+                 "\nPlease try again, using 'sudo'. Exiting.")
+        # Check for basic utilities
+        smm = SoftwareManager()
+        detected_distro = distro.detect()
+        packages = ['gcc', 'make']
+
+        if detected_distro.name == "SuSE":
+            packages.extend(['git-core'])
```
apahim, line 58: Well, if it's only one package, you don't need to create a list. You can use `packages.append('git-core')` instead.

apahim:
looks good.

PraveenPenguin:
@apahim oh ok , sure will keep in mind for new test cases .

Thanks

PR #135, opened 2016-08-12
Software Raid test case

vrbagalkote:
Signed-off-by: Venkat R B vrbagal1@linux.vnet.ibm.com

### DIFF
--- a/io/disk/softwareraid.py
+++ b/io/disk/softwareraid.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2016 IBM
+# Author: Venkat Rao B <vrbagal1@linux.vnet.ibm.com>
+
+
+"""
+RAID devices are virtual devices created from two or more real block devices.
+This allows multiple devices (typically disk drives or partitions thereof) to
+be combined into a single device to hold (for example) a single filesystem.
+Some RAID levels include redundancy and so can survive some degree of device
+failure.
+"""
+
+import re
+from avocado import Test
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import process
+from avocado import main
+
+
+class SoftwareRaid(Test):
+
+    """
+    This scripts create, assembles and stops md device, using mdadm tool
+    """
+
+    def setUp(self):
+
+        """
+        Checking if the required packages are installed,
+        if not found packages will be installed
+        """
+
+        smm = SoftwareManager()
+        if not smm.check_installed("mdadm"):
+            self.skip("Mdadm must be installed before continuing the test")
+            if SoftwareManager().install("mdadm") is False:
+                self.skip("Unable to install mdadm")
+        cmd = "mdadm -V"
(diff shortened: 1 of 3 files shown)


io/disk/softwareraid.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        Checking if the required packages are installed,
+        if not found packages will be installed
+        """
+
+        smm = SoftwareManager()
+        if not smm.check_installed("mdadm"):
+            self.skip("Mdadm must be installed before continuing the test")
+            if SoftwareManager().install("mdadm") is False:
+                self.skip("Unable to install mdadm")
```
PraveenPenguin, line 51: this code `if SoftwareManager().install("mdadm") is False:` never be going to execute as it check madm is installed if it not installed  it skip the test

PraveenPenguin, line 51: better to use self.error instead of self.skip

because in `self.error i`t flash message as `TestSetupFail:` but in `self.skip` it is TestSkipError

PraveenPenguin, line 51: here seems me repetition of code but  No strong opinion about it. rather than  use of  multiple code for package check it can be 
`if not sm.check_installed('mdadm') and not  sm.install('mdadm'):`  
so please consider it

io/disk/softwareraid.py
```diff
@@ -0,0 +1,112 @@
...
+                          to create Raid5/10")
+        if self.raidlevel == '6':
+            if self.disk_count < 3:
+                self.skip("Minimum of three disk are required to create Raid6")
+
+    def test_createraid(self):
+
+        """
+        Creatingstoping and assemblying the raid
+        """
```
PraveenPenguin, line 81: I don't follow the meaning of this. can you please  formate or tell differently?

io/disk/softwareraid.py
```diff
@@ -0,0 +1,112 @@
...
+            cmd = "losetup --find"
+            varstr = process.system_output(cmd, ignore_status=True, shell=True)
+            var = int(re.match(r'(/dev/loop)(\d+)', varstr).group(2))
+            cmd = "dd if=/dev/zero of=/file%d bs=1k count=100" % var
+            process.system(cmd, ignore_status=True, shell=True)
+            cmd = "losetup /dev/loop%d /file%d" % (var, var)
+            process.system(cmd, ignore_status=True, shell=True)
+            self.loop += "/dev/loop%d" % var
+        self.disk = self.params.get('disk', default=self.loop)
+        self.disk = self.disk.strip(' ')
```
PraveenPenguin, line 66: why not this  `self.disk = self.params.get('disk', default=self.loop).strip(' ')` ?

io/disk/softwareraid.py
vrbagalkote, line 51: Yes, you are right. I have made the changes as proposed.

smm = SoftwareManager()
        if smm.check_installed("mdadm") is False:
            print("Mdadm is not installed")
            if SoftwareManager().install("mdadm") is False:
                self.skip("Unable to install mdadm")

io/disk/softwareraid.py
vrbagalkote, line 81: Here, I am checking for the raid level and also number of disk passed in the yaml file, if any.
As each raid level needs minimum number of disks to be created, So I am checking if a particular raid can be configured on the given number of drives.

In softwareraid, for raid5, 10 minimum number of drives required are 3, and for raid6 is 4

io/disk/softwareraid.py
vrbagalkote, line 66: Yes, I have taken care of this in the softwareraid_v2.py.

io/disk/softwareraid.py
PraveenPenguin, line 81: @vrbagalkote thanks a lot for info , just want to  format this "Creatingstoping and assemblying the raid" than it can be meaning full .

Thanks for your support

vrbagalkote:
Created a new pull request #141 addressing all the above comments.

PR #136, opened 2016-08-13
Adding NVMe cli tests

narasimhan-v:

### DIFF
--- a/io/disk/ssd/nvmetest.py
+++ b/io/disk/ssd/nvmetest.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+"""
+NVM-Express user space tooling for Linux, which handles NVMe devices.
+This Suite creates and formats a namespace, reads and writes on it
+using nvme cli.
+"""
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import process
+import avocado.utils.git as git
+
+
+class NVMeTest(Test):
+
+    """
+    NVM-Express user space tooling for Linux, which handles NVMe devices.
+
+    :param device: Name of the nvme device
+    """
+
+    def setUp(self):
+        """
+        Build 'nvme-cli' and setup the device.
+        """
+        nvme_path = os.path.join(self.srcdir, 'nvme')
+        nvme_link = 'https://github.com/linux-nvme/nvme-cli.git'
+        git.get_repo(nvme_link, destination_dir=nvme_path)
+        build.make(nvme_path)
+        build.make(nvme_path, extra_args='install')
+        self.device = self.params.get('device', default='/dev/nvme0')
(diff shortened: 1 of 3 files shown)


io/disk/ssd/nvmetest.py
```diff
@@ -0,0 +1,114 @@
...
+        nvme_path = os.path.join(self.srcdir, 'nvme')
+        nvme_link = 'https://github.com/linux-nvme/nvme-cli.git'
+        git.get_repo(nvme_link, destination_dir=nvme_path)
+        build.make(nvme_path)
+        build.make(nvme_path, extra_args='install')
+        self.device = self.params.get('device', default='/dev/nvme0')
+        # cmd = "ls -l %sn* | sort | awk '{print $NF}'" % self.device
+        # self.id_ns = process.system_output(cmd, shell=True).strip('\n'):
+        # if 'No such file or directory' in self.id_ns:
+        #     self.id_ns = self.device + 'n1'
```
PraveenPenguin, line 52: why these commented code  , seems me clean up is require

io/disk/ssd/nvmetest.py
```diff
@@ -0,0 +1,114 @@
...
+    def create_namespace(self):
+        """
+        Creates namespace on the device.
+        """
+        cmd = "nvme show-regs %s | grep version | awk '{print $NF}'" \
+            % self.device
+        if int(process.system_output(cmd, shell=True).strip('\n')) < 10200:
+            return self.device + 'n1'
+        # cmd = 'nvme create-ns %s' % self.device
+        # process.run(cmd, shell=True)
```
PraveenPenguin, line 72: same here as well as commented code

io/disk/ssd/nvmetest.py
```diff
@@ -0,0 +1,114 @@
...
+        print self.device, self.id_ns, self.lba, self.format_size
+
+    def create_namespace(self):
+        """
+        Creates namespace on the device.
+        """
+        cmd = "nvme show-regs %s | grep version | awk '{print $NF}'" \
+            % self.device
+        if int(process.system_output(cmd, shell=True).strip('\n')) < 10200:
+            return self.device + 'n1'
```
PraveenPenguin, line 70: can you please help me for understanding  requirement of this  `if statement`as i figure out in out side of if also    return same value 

`return self.device + 'n1'`

io/disk/ssd/nvmetest.py
```diff
@@ -0,0 +1,114 @@
...
+        self.id_ns = self.create_namespace()
+        self.format_namespace()
+        cmd = "nvme id-ns %s | grep 'in use' | awk '{print $5}' | \
+            awk -F':' '{print $NF}'" % self.id_ns
+        self.format_size = process.system_output(cmd, shell=True).strip('\n')
+        self.format_size = pow(2, int(self.format_size))
+        cmd = "nvme id-ns %s | grep 'in use' | awk '{print $2}'" % self.id_ns
+        self.lba = process.system_output(cmd, shell=True).strip('\n')
+        print self.device, self.id_ns, self.lba, self.format_size
+
```
PraveenPenguin, line 62: wondering about print statement as stdout (job log)  will having all run info  or can be used self.log.info()

io/disk/ssd/nvmetest.py
narasimhan-v, line 70: That is because I am yet to implement the determination of actual self.id_ns value, for which I need a latest adapter.
Since I do not have that yet, this is a placeholder.
I have mentioned this in TODO

PraveenPenguin, line 70: @narasimhan-v  ok thanks

narasimhan-v:
@PraveenPenguin Implemented the suggested changes, and created PR #138

PraveenPenguin:
@narasimhan-v thanks will check and try to run in my environment ,

one question i did not see any info in which system it going to run  like mentioned in README 
i think it should run only a  system where NVMe Device available but what about other  system , can  we need to handle that situation in test ? or have info in README  
Please advise
Regards
Praveen

narasimhan-v:
@PraveenPenguin Yes, it can run only on the system with nvme devices.
It is mentioned in README file.

PR #143, opened 2016-08-16
added xfstests Test case[v2]

PraveenPenguin:
Signed-off-by: Praveen K Pandey praveen@linux.vnet.ibm.com

v1 #105 

`v2 :  added skip logic in setup`

Regards 
Praveen

### DIFF
--- a/fs/xfstests.py.data/README
+++ b/fs/xfstests.py.data/README
@@ -1,0 +1,43 @@
+This is a simple wrapper for running xfstests inside avocado. The steps to get
+started are really simple:
+
+1) Edit the configuration variables on the yaml file as well environment variable
+
+1.1) The variables 'TEST_DEV' and 'TEST_DIR' are mandatory and should be set to
+     a block device path and mount point path, respectively, that will be used
+     *exclusively* for xfstests. It must have the filesystem of your choice
+     previously created.
+     ex : export TEST_DEV =/dev/vda1 , export TEST_DIR=/mnt
+
+     DO NOT USE A BLOCK DEVICE WITH IMPORTANT DATA!!!
+
+1.2) Set the range of tests you want to run setting the TEST_RANGE variable.
+     Please notice that python's range() function may not work as you expect,
+     that is, if you want a range from 0-255, use: range(0, 256)
+
+General notes
+-------------
+
+* As autotest includes a setup phase for client tests, this step is encapsulated
+in a dummy xfstests number 000.
+
+* XFS utilities, system libraries and header files are checked early, before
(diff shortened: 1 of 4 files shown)


apahim:
Can you please provide the job.log for this version?

fs/xfstests.py.data/README
```diff
@@ -0,0 +1,43 @@
...
+     a block device path and mount point path, respectively, that will be used
+     *exclusively* for xfstests. It must have the filesystem of your choice
+     previously created.
+     ex : export TEST_DEV =/dev/vda1 , export TEST_DIR=/mnt
+
+     DO NOT USE A BLOCK DEVICE WITH IMPORTANT DATA!!!
+
+1.2) Set the range of tests you want to run setting the TEST_RANGE variable.
+     Please notice that python's range() function may not work as you expect,
+     that is, if you want a range from 0-255, use: range(0, 256)
```
apahim, line 16: What does the ptyhon `range()` have to do with your test? Is this just a heads up for those setting this env variable using the `range()`?

fs/xfstests.py.data/README
```diff
@@ -0,0 +1,43 @@
+This is a simple wrapper for running xfstests inside avocado. The steps to get
+started are really simple:
+
+1) Edit the configuration variables on the yaml file as well environment variable
```
apahim, line 4: Maybe you should separate what is env var from what is set in yaml file. Anyway, I don't see any clue about what goes in the yaml file.

fs/xfstests.py.data/README
apahim, line 16: Maybe an example here would help.

PraveenPenguin, line 16: here range means test no in the range if we pass 001,014 than test will all run test  001 to 014 test no it can be given in yaml file

fs/xfstests.py.data/README
PraveenPenguin, line 4: @apahim while start the test  env variable need to set as "TESTDIR and TESTDEV using export"  and configuration value will  be like test_number and skip_dangerous

PraveenPenguin:
will attach job file as soon as possible .

fs/xfstests.py.data/README
PraveenPenguin, line 16: @apahim  thanks a lot to pointing me in this , oh sorry cheked in wrong README file , TEST_RANGE still in TODO , let me chek in right file as well as upload job log as well .

Thanks 
Praveen

PraveenPenguin:
@apahim raise a new PR #145  as i figure out that OSS git is not working today and check in modified README file Please have a look , 

Thanks

